PR #7, opened 2023-08-12
death_anims.cpp: Use appropriate buffer sizes

clayne:
* rnd_motion::setup: Due to a lack of safety in _CopyVal (fixed separately) the string64 buffer can easily be overrun. Extend this to 1024 to be in line with the expected max length of a line. Also add an assertion and clean up the magic number stuff (1023).

### DIFF
--- a/src/xrGame/death_anims.cpp
+++ b/src/xrGame/death_anims.cpp
@@ -19,10 +19,11 @@ rnd_motion* rnd_motion::setup(IKinematicsAnimated* k, LPCSTR s)
 	VERIFY(k);
 	VERIFY(s);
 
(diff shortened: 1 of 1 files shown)


clayne:
Related: https://github.com/themrdemonized/xray-monolith/pull/6

clayne:
This was discovered by a build with stack safety checking enabling and a mod that errantly assumed it could use `/` separated values for `random_death_animations`. The code was trying to copy a 79 byte value into a 64 byte buffer due to a problem with `_CopyVal` outlined here: https://github.com/themrdemonized/xray-monolith/pull/6

image

image

PR #16, opened 2023-10-19
Add a console command to toggle mouse wheel zoom direction

mbehm:
One of my pet peeves with Stalker games has been the mouse wheel zoom being IMO inverted. Thanks to demonized providing an easy to build repo and a perfect example with the mouse wheel weapon switch toggle, here's a simple switch to toggle the zoom direction for both weapons and PDA map.

`mouse_wheel_invert_zoom 1` (default) Mouse wheel up zooms out, mouse wheel down zooms in, set by default to match original behavior (which I think is inverted).

`mouse_wheel_invert_zoom 0` Mouse wheel up zooms in, mouse wheel down zooms out.


mbehm:
On a side note while this PR aligns with the way the mouse wheel weapon switch has been implemented, I'd be highly in favor of using `g_` prefix for globals like this.

themrdemonized:
Thanks, i will consider changing naming of cvars in the future

PR #20, opened 2024-01-17
Increased lua_gcstep default to 400 and allow console editing

MagielBruntink:
Hi @themrdemonized,

This is an experiment with changed lua_gcstep variable. I think by default it is set too low in xray at the moment. The current default 100 is actually half the step size of lua/luajit 's GC, which is 200. I'm playtesting with 400, with better results (fewer stutters) than on the default 100.

It seems also a step size of 1000 is too much - leading to average FPS decrease.

### DIFF
--- a/src/xrGame/Level.cpp
+++ b/src/xrGame/Level.cpp
@@ -764,7 +764,7 @@ void CLevel::OnFrame()
 	}
 }
 
-int psLUA_GCSTEP = 100;
+int psLUA_GCSTEP = 400;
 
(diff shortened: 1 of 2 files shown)


MagielBruntink:
Relevant Lua 5.1 docs: https://fxcodebase.com/bin/sdk2beta/help/lua/lua_gc.html

MagielBruntink:
Oh cool, merged already :-) Playtests so far have been really much smoother. Hope it's the same for others!

PR #22, opened 2024-01-20
Disable ogg-comment logging in MASTER_GOLD.

MagielBruntink:


themrdemonized:
What is the point of this change?

MagielBruntink:
Cleaning up the console and the log file. These ogg comment warnings really mess up both of those.

themrdemonized:
Those are pretty useful for modders, and using that directive will remove them entirely. If you want them to be displayed only in debug mode, instead of directive use `strstr(Core.Params, "-dbg")`

MagielBruntink:
Closing, new approach in https://github.com/themrdemonized/xray-monolith/pull/23

PR #28, opened 2024-03-18
HeatVision flares fix + hot sun 

vegeta1k95:
1) Fix issue when lens flares - from sun, etc - were rendered on top of infrared image.
2) Rendering sun/moon into heat RT - makes it possible for those to be displayed hot in infrared.


themrdemonized:
thanks

PR #29, opened 2024-03-20
Add npc_on_item_before_pickup callback

Tosox:
Did some play testing and didn't stumble upon errors.
The callback doesn't fire when the NPC receives an item (when spawning; by trading; delayed mechanic weapon repairs), but when the NPC walks over the item which is then transferred to its inventory. To my knowledge NPCs can only pickup weapons (including melee weapons) that way.


themrdemonized:
thanks

PR #32, opened 2024-03-25
Add optional "blowout_disable_idle" option for anomalies

vegeta1k95:
Setting option "blowout_disable_idle" to true (false by default) in a config will make anomaly temporarily turn off idle particles when anomaly's blowout is triggered, particles are returned upon return to idle state. 

Allows more fine-tuning of anomalies' visuals for modders.


themrdemonized:
thanks

PR #33, opened 2024-03-30
Fixed gravity anomalies not always playing particles/sounds of body tearing

vegeta1k95:
1) Fix the case, when gravity anomalies would not play particles and sounds of body tearing of mutants/NPC who died in anomalies.

**Explanation:**
There are two particles sections in gravity anomalies configs:
`tearing_particles` and `torn_particles`. 

- The `tearing_particles` are played on the (closest to root) **particle bone** of ORIGINAL object - corpse of NPC or mutant.
- The `torn_particles` are played on all bones of DESTROYED object - this is a skeleton (ph_skeleton_object) which is spawned after anomaly rips the original body apart.

**First:** `torn_particles` were never really played because of some weird check against `CEntityAlive`, which `ph_skeleton_object` isn't. I removed that check, so now `torn_particles` are played properly.

**Second:** `tearing_particles` were only played if object in question has properly defined `particle_bones` (bones which can hold and play particles) inside of their configs. This lead to the situation that some mutants and human NPCs could not play particles because there was no defined bone to attach to. (This is also partially related to the fact that some mutants could not be dismembered - they are considered as solid object with one bone)

**Fix:** I totally detached `tearing_particles` effect from original object's bones. Instead I create auto-removing particle effect at the position of the original object, so that object in question does not have to have properly setup `particle_bones` - particles will just play at the object's position at the time of explosion.

These are the main changes in `CMincer::NotificateDestroy(...)` and `CTeleWhirlwindObject::destroy_object(...)`

2) I did some clean up and prettify classes `CMincer` and `CTeleWhirlwind` - no logic changes, just some prettifying for variables, function names, etc.

3) Added optional parameter `tele_height_fixed` (default `true`). If `false` telekinesis of gravity anomalies will not hard-stop objects at the peak point (`tele_height`) - they would fluctuate around it instead.

### DIFF
--- a/src/xrGame/Mincer.cpp
+++ b/src/xrGame/Mincer.cpp
@@ -42,13 +42,17 @@ void CMincer::Load(LPCSTR section)
 {
 	inherited::Load(section);
 
-	m_telekinetics.set_destroing_particles(shared_str(pSettings->r_string(section, "tearing_particles")));
-	m_telekinetics.set_throw_power(pSettings->r_float(section, "throw_out_impulse"));
-	m_torn_particles = pSettings->r_string(section, "torn_particles");
-	m_tearing_sound.create(pSettings->r_string(section, "body_tearing_sound"), st_Effect, sg_SourceType);
-	m_fActorBlowoutRadiusPercent = pSettings->r_float(section, "actor_blowout_radius_percent");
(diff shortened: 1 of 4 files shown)


themrdemonized:
thanks

PR #34, opened 2024-03-31
Demo_Record update

strangerism:
- added to DemoRecord extra keybindings to support pc/laptops with no numpad
- added new console command (demo_record_return_ctrl_inputs) to launch DemoRecord with the ability for client scripts to receive certain keystrokes and react to it while DemoRecord is still running


themrdemonized:
thanks

PR #35, opened 2024-04-02
Fix some hud particles alignment (smoke)

vegeta1k95:
This should fix the problem where weapon muzzle flash and smoke particles were wrongly aligned wrt each other.


themrdemonized:
thanks

PR #36, opened 2024-04-04
Editable variables that allow weapon position adjustements that don't require aim tweaks

DaimeneX:
To do: Debug mode weapon position modifier. At current moment just adding the new value results in the game treating base_hud_offsets as default value which isn't ideal.


themrdemonized:
thanks

PR #38, opened 2024-04-06
base_hud_offsets handling for PDA and Devices

DaimeneX:
Added code to allow PDA and Devices to handle hud_base_offsets


DaimeneX:
You'd think it would've been as simple as before, but it turns out this code nullified base_hud_offsets and removing it made the aim jump between one state and the other.
I added tweaked weapon.cpp and weapon.h code that allows for smooth transitions between states. As a result the aim in and out is a bit smoother as well.
`curr_offs.mul(m_fZoomfactor);
curr_rot.mul(m_fZoomfactor);`

PR #40, opened 2024-04-12
DemoRecord. adds second console parameter, reposition camera at start

strangerism:
First of all seems my commits changed the whole files but in reality I only changed few lines.

I think it's because my local git added unix line endings. I had the `git config --global core.autocrlf` set to false temporarily because prior to this I was working on something else that needed the unix line endings. My bad I forgot to change back the git config.
An explanation can be found here in the first response https://stackoverflow.com/questions/2825428/why-should-i-use-core-autocrlf-true-in-git

Anyway I could try to revert the head to a previous commit but I worry is going to even make more mess in the timeline, So if it's ok for you to take this PR then fine otherwise I will try to clean up my commits 

The change:

- added a second optional parameter when launching demo_record e.g. demo_record_return_ctrl_inputs <filename> <boundary>
- changed the camera position at start, to behind the actor instead of being inside

So let me know what to do, I personally hate this kind of commit blunder unfortunately I could not see it until it pushed to github.
Regardless of everything, the project builds fine and the exes too.

Waiting for your advice

### DIFF
--- a/src/xrEngine/FDemoRecord.cpp
+++ b/src/xrEngine/FDemoRecord.cpp
@@ -102,11 +102,13 @@ CDemoRecord::CDemoRecord(const char* name, float life_time) : CEffectorCam(cefDe
 
 		// Set the camera position to the actor's position
 		m_Actor_Position.set(m_Camera.c.x,m_Camera.c.y,m_Camera.c.z);
+
 		// Move the camera a bit in front of the actor
 		float distanceInFrontOfActor = 3.0f; // Set this to the desired distance
 		Fvector cameraOffset = m_Camera.k;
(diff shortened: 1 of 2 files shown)


strangerism:
image
console_commands.cpp

strangerism:
image
FDemoRecord.ccp

strangerism:
ok maybe reject this Pull Request. I try to commit the files as they are but with no linux line endings

strangerism:
so it turns out that the diff on github shows the full file changed because it added a bunch of white spaces my commits 
image

you can still see the real change if you select "ingore whitespaces" in the github toolbar
image

themrdemonized:
can you make it so the behind the actor camera start only works with your new command?

PR #41, opened 2024-04-13
DemoRecord: adds second console parameter, refactored constructor to pass command flag, reposition camera at start

strangerism:
Refactored per your suggestion

The change:

- added a second optional parameter when launching demo_record e.g. demo_record_return_ctrl_inputs
- refactored the class constructor the alternate mode flag (return_ctrl_inputs) and handle carema reposition based on that
- changed the camera position at start, to behind the actor instead of being inside, only when return_ctrl_inputs is true

### DIFF
--- a/src/xrEngine/FDemoRecord.cpp
+++ b/src/xrEngine/FDemoRecord.cpp
@@ -102,11 +102,13 @@ CDemoRecord::CDemoRecord(const char* name, float life_time) : CEffectorCam(cefDe
 
 		// Set the camera position to the actor's position
 		m_Actor_Position.set(m_Camera.c.x,m_Camera.c.y,m_Camera.c.z);
+
 		// Move the camera a bit in front of the actor
 		float distanceInFrontOfActor = 3.0f; // Set this to the desired distance
 		Fvector cameraOffset = m_Camera.k;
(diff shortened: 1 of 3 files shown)


strangerism:
I could not turn the camera 180 so I had to do with moving the camera behind and call it done. 
This is probably my last edit on this thing, unless I found a way to do it.
Another thing I would have loved is to give the camera effector some collision checks (player and environment) but I have no clue if this is possible.

themrdemonized:
i pushed your changes via separate commit to clean up the merge mess.

strangerism:
ok thanks, sorry about that. I still don't know the reason my fork or my loca git did that. Do you think that after your commit, a resynch is going to fix it on my end as well?

themrdemonized:
yea, just sync forks, when it asks for conflicts, tell to use my version

strangerism:
```
PS E:\dev\anomaly\xray-monolith> gh repo sync strangerism/xray-monolith -b all-in-one-vs2022-wpo                                                            
can't sync because there are diverging changes; use `--force` to overwrite the destination branch
```
I guess I overwrite?

themrdemonized:
yea

PR #42, opened 2024-04-14
Poltergeists and Burers throw corpses

Tosox:
I had to modify `telekinetic_objects.cpp` since it appears that the height at which objects levitate above ground during the attack phase correlates with the object's physics shell elements count (`object->m_pPhysicsShell->Elements().size() ^ 2 * strength`). Therefore, objects with fewer elements like weapons hover at a regular height, while those with more elements like bodies hover very high in the air. To resolve this, I've capped the element count at a maximum of 5.

The proposed code alteration enables both poltergeists and burers to utilize their telekinetic attack to throw corpses of stalkers and mutants. By default, this feature is disabled but can be enabled through either the options menu or via command.

Burers preview: https://youtu.be/DrYh5UuXoRU
Poltergeists preview: https://youtu.be/Outy2Fq6OvQ

### DIFF
--- a/gamedata/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -118,4 +118,8 @@
 	<string id="ui_mm_modded_exes_freelook_cam_limit">
 		<text>Freelook angle limit (freelook_cam_limit)</text>
 	</string>
+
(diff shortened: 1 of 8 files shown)


Tosox:
elem_size maybe requires more clamping

Tosox:
I think this PR is ready now. The feature is disabled by default so it shouldn't break anything.

themrdemonized:
thanks

PR #43, opened 2024-04-21
Added silencer heat glowing functionality (optional)

vegeta1k95:
Adds an optional and configurable texture glowing for silencers after consecutive shots, visually looks like overheating.

This just enables to do so from engine side. Shader part will be provided later as an addon (without it nothing will happen).

Code changes are very similar to those for HeatVision.


themrdemonized:
thanks

PR #44, opened 2024-05-02
Added `GetWindow()` method

vegeta1k95:
A method to access main HUD's `CUIWindow` instance.


themrdemonized:
I would still discourage use of this for adding statics or other elements. The proper way would be deriving from CUIScriptWnd class and using get_hud():AddDialogToRender

```
HUD = UICompanionList()
get_hud():AddDialogToRender(HUD)

function actor_on_net_destroy()
	if HUD ~= nil then
		get_hud():RemoveDialogToRender(HUD)
		HUD = nil
	end
end

class "UICompanionList" (CUIScriptWnd)
function UICompanionList:__init() super()
...
```

PR #45, opened 2024-05-04
adds a new console command freeze_time

strangerism:
Adds a new console command that allows to freeze the alife time e.g. time_factor [0,1] but the sounds can still play.

While the time is frozen, any current playing sounds, from alife scripts or soundtracks will keep playing until their end  while looped sounds (anomalies fx) will play normally. You can also play new sounds in this frozen state from scripts

Check video to see how this command is being used https://discord.com/channels/961738167139049472/961738168162484317/1236213384735752222


strangerism:
wait on this one. I need to check something

strangerism:
I made a fix to this code, will create a new one

PR #47, opened 2024-05-09
Added custom grenade explode effector

VodoXleb:
When grenade explodes near player, explode effector is played.
I changed the code to make it customizable to each grenade
e. g.
if grenade section has:
explode_effector_sect_name = $no_effector
then there will be no effector.
if it has any effector sect then it will be played


VodoXleb:
With this we can disable earthquakes caused by smoke grenades for example

themrdemonized:
thanks

PR #48, opened 2024-05-11
feature(wpn-zoom-type-changed-callback): Add Callback On Weapon Zoom Type Changed

nltp-ashes:
Added a callback that triggers when actor changes zoom type. The callback is fired with three arguments.

- `obj` : `game_object` - the weapon object for which the zoom type was changed
- `previous` : `number` - the previous zoom type, before the change (`0` = normal, `1` = alt, `2` = gl)
- `current` : `number` - the current zoom type, after the change (`0` = normal, `1` = alt, `2` = gl)


themrdemonized:
thanks

PR #55, opened 2024-06-12
Fix PhraseDialog.cpp

VodoXleb:
Trying to fix Dialog crashes occuring after 20.05 update of PhraseDialog.cpp by GhenTuong.(It's his fix)


themrdemonized:
Pushed via separate commit to preserve devs comments

PR #57, opened 2024-07-19
Add callbacks for getting artefact_count of outfit and getting belt_size in inventory.

VodoXleb:
Now possible to intercept these calls and put our own values instead.
For example we can add artefact slots for backpack with these changes or easily add additional slots to outfits without exiting the game.
Changed CustomOutfit and Inventory.cpp.


themrdemonized:
I'll checkout later, thanks

PR #59, opened 2024-08-03
Add viewport_near console command to change camera near value

VodoXleb:
While mostly unused, can be useful to first person death and in similar cases


VodoXleb:
will be helpful to avoid clipping through floor after dying

VodoXleb:
hmm, guess you already managed that

VodoXleb:
anyway I would like to add this

PR #62, opened 2024-08-24
3D Shader scopes rendering adjustments

Redotix:
The purpose of these edits is to allow 3DSS to control the render order of its lenses and provide missing data to fix rendering issues.

Additions:
- custom shader flags for 3DSS lenses with a custom render order
- a render phase for rendering the scope reticle 
- a render target that samples the z buffer of the scene before hud rendering begins


themrdemonized:
thanks

PR #65, opened 2024-09-04
Fix HDR10 issues with HUD samplers and MSAA setting

deggua:
- Fixed the HUD samplers for `hud_default.s` and `hud_cursor.s`, they were using WRAP'd sampling instead of CLAMP'd so would break with certain texture configurations expecting CLAMP'd sampling
- Added `clamp()` method support to Lua shader configuration files to support this, there are others that might be useful which are commented out, but I'm hesitant to just implement them without anything using them since some of them may not work correctly, they must have been commented out for a reason, but at least the `clamp` option seems to work correctly
- Force disable MSAA in DX11 if HDR is enabled since it would break the renderer if both were enabled simultaneously, this may be confusing for users without any text in the HDR option saying it's incompatible with MSAA, is there a way to provide a tooltip or something to the Modded Exe's menu that could describe this?


themrdemonized:
Thanks, i'll add the tooltip warning in Modded Exe's options

PR #66, opened 2024-09-05
Fix for D3D management issues in `dx10HW.cpp`

deggua:
The renderer was unable to resize buffers after a fullscreen state change (required for flip discard swap chain) because the depth and color targets (views) still had references to them, now they're released before `ResizeBuffers` is called. Also removed calls to `ResizeBuffers` and `UpdateViews` in `DestroyDevice` since resizing and re-creating the views is unnecessary if the entire swap chain will be destroyed and re-created by `CreateDevice`.

`Reset` call in `CreateDevice` handles all the resizing and updating required for the default render target to match the swap chain resolution. Before, when only calling `UpdateViews`, the default color target could be a different resolution than the swap chain, so the depth buffer would have a resolution mismatch with the color target, which would result in a black screen until the user alt-tabbed out and back in.


themrdemonized:
Thanks

PR #70, opened 2024-09-13
Allow rifle slot to be single-handed

Y3sMan:


Y3sMan:
Simple change to allow the rifle weapon slot (slot 2) to also allow single_handed-ness. Additionally, add a check after the weapon type casting.

themrdemonized:
Hi, what's the purpose of this change. Is it to make possible to use detectors with rifles?

Y3sMan:
> Hi, what's the purpose of this change. Is it to make possible to use detectors with rifles?

Sorry about that, yes it is. Now the weapon does not need to be in the pistol slot, it can simply be in the main rifle slot and still allow the detector to be pulled out.

themrdemonized:
Please make it so it can be toggleable via console, with disabled as default

Y3sMan:
"toggleable via console"
Each weapon still needs to have the line `single_handed = 1` under their ltx config, the feature is disabled by default. Or do you want something else to be toggeable?

themrdemonized:
alright, i thought it just enables it for any weapon in the rifle slot

themrdemonized:
aren't `iitm.IsSingleHanded()` is commented in your code?, it comes after //

Y3sMan:
The bolt does break it, I'll fix that. But the check happens at `W->IsSinglehanded()`, after casting to the Weapon type.

Y3sMan:
Alright, I added a dltx config setting `single_handed = 1` on the bolt.

themrdemonized:
I'll check it out later, thanks

SevenChalices:
There are 1-2 VERY useful add-ons (can't imagine to play without it after months of enjoying it), the first one is afaik well known, it lets you put melee weapons, handguns (pistols & revolvers) into your binocular and/or knife slot (and vice versa). Imo an absolute must even for vanilla-lovers.

https://www.moddb.com/mods/stalker-anomaly/addons/knife-and-pistol-in-binoculars-slot
(install with Mod Organizer for bad operating systems or if installing manually, compare conflicting files with any add-on that affects your inventory UI (f.e. that one which displays sleepyness-thirst-hunger, but that's really not difficult).

The second one is afaik known much less (as it was released 3 years after the first add-on) - it can be seen as an improvement of the former mentioned add-on. With this, you can do the same (use knife/binoc.-slots) but instead of "re-slotting" only pistols it lets you to put most "compact firearms" (sawn-offs, compact SMGs like Skorpion, Kiparis, MP7...)
I am not entirely sure, but this add-on automatically adds that "single_handed = true" for all those weapons. In one word: eto prosto pizdec - but without making you overpowered.

I am not sure, but it can be that in the conflicting configs/ui/*.xml file(s) there's already a "patch" integrated for the first add-on. Just search the file for some (XML-style-)comments.

https://www.moddb.com/mods/stalker-anomaly/addons/dltx-machine-pistols-sawn-off-in-sidearm-slot

Hope I could help you - and to save TheMrDemonized to be annoyed by other users than me ;) (just kidding)

Y3sMan:
Looking back, I very obviously did not explain this PR very well, that's my bad. I'll reply here then addendum my PR.
The short of it is Anomaly will not allow you to define weapons solely as "single_handed;" weapons can only be single-handedly if they are `slot = 2` weapons or they are in lower slot.
So, this means, if I want to build a mod that, say, allows the detector to be pulled out with every weapon, I'd allow every weapon to be single_handed, but in the main branch as is this means every gun would also the fit into the pistol/knife slot as well! Not good. The PR would allow `slot = 3` rifles, shotguns, other general "not-small-guns" to be single_handed without letting players now have 4 sniper rifles equipped at all times.
Hope this helps :)

PR #71, opened 2024-09-14
Fix HDR10 rendering

deggua:
This fixes a bunch of stuff that was wrong before
- Tonemapping still needs to be applied, but I delay doing it until the end of the render pipeline (in the `postprocess` shaders)
- Tonemapped values are properly scaled by the display max luminance, this fixes whites being blown out, etc.
- Color space conversion was incorrect before, should be correct now. Had to convert to target color space before tonemapping, then to display colorspace after to limit the color gamut correctly (since Windows expects Rec.2020)
- Added parameters for adjusting the final output
  - Tonemapper (just implemented a bunch of options, but I think the ACES ones probably look the best)
  - Tonemapping mode (color or luminance, luminance prevents hue shift, but looks desaturated and flat, color is preferable imo)
  - PDA intensity (the PDA rendering is tricky, just added an extra scalar if it's too bright)
  - Exposure (to brighten the image if it's too dark after tonemapping)
  - Contrast (to adjust contrast, some of the tonemappers have really exaggerated contrast)
  - Contrast Middle-Gray (fixed point for contrast adjustment, lower values cause contrast adjustment to affect brighter colors more than darker colors, higher values do the opposite)
  - Saturation (to adjust saturation)
- DOF render target is upgraded to F16 since it renders directly into final output, not doing this causes the HDR values to be clipped before they get to the combine pass
- Renamed DCI-P3 colorspace to P3-D65 (they seem to be used interchangeably but have different whitepoints, I think P3-D65 is used for HDR on monitors and is what the HDR analysis shaders I've been using seem to target)
- No longer need to modify the `effects_flare` or `effects_sun` shaders in HDR, I think the sky blowout was just due to not tonemapping before applying the ST2084 curve (basically just gamma curve for HDR)

I've been using the HDR analysis shader from https://github.com/EndlesslyFlowering/ReShade_HDR_shaders to check stuff like color space output and max nits is correct, and it seems like everything is. Only thing that might need adjustment is the contrast slider/parameters. It works, but large values have weird effects.

The Russian translations are machine translations so feel free to change them if they're bad.

Also I wrote a small tool to read HDR info about monitors on Windows. Might be useful for people tuning the HDR10 parameters because my monitor actually reports higher peak luminance than the mode in the monitor OSD indicates.
https://github.com/deggua/hdr-query


themrdemonized:
Thanks

PR #83, opened 2024-10-09
Aim factor for hud_fov

damoldavskiy:
Adds parameter which lowers hud_fov in ads. Useful for 3D scopes. Default off


themrdemonized:
Next time you do a PR, describe it more thoroughly please.

For example: Added `hud_fov_aim_factor` console parameter which  which lowers hud_fov in ads. Useful for 3D scopes. Default is 0

PR #112, opened 2024-12-08
A handful of miscellaneous changes

etapomom:
- Added console commands:
	1. fix_avelocity_spread -> Bool; Fixes moving the camera not affecting accuracy.
	2. apply_pdm_to_ads -> Bool; Makes it so that PDM values from the weapon are properly applied when aiming down sights.
	3. smooth_ads_transition -> Bool; Makes it so that going in and out of ADS smoothly changes the accuracy bonus instead of changing only at full ADS.
	4. allow_silencer_hide_tracer -> Bool; Optionally disables silencers being able to hide tracers completely. (Defaults to true)

- Added silenced_tracers (bool) field to weapons sections. If the ammo type can be hidden when using a silenced weapon, setting this field to true makes the tracer visible again.
- Added GetZoomRotateTime() and SetZoomRotateTime(float val).
- Added senderweapon to the table used for level.add_bullet() to improve compatibility with modding. Defaults to sender.
- Added bkeep_speed to SetActorPosition().
- Added SetMovementSpeed(). Exported as set_movement_speed(vector vel) for scripting.

### DIFF
--- a/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -77,6 +77,13 @@
 		<text>Particle update rate (1 - 30hz, 0.5 - 60hz, etc) (particle_update_mod)</text>
 	</string>
 
+	<string id="ui_mm_modded_exes_render_short_tracers">
+		<text>Cap and render tracers to their minimum length (render_short_tracers)</text>
+	</string>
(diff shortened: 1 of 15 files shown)


themrdemonized:
Please add new exports into lua_help_ex.script and add new console parameters into modded exes options ui in `ui_mm_modded_exes.script`

etapomom:
done, also added my previous changes to those files.

themrdemonized:
Thanks

PR #122, opened 2025-01-18
Stutter improvement, disable running 'net_Relcase' on scripts

TIHan:
There is a script function called `net_Relcase` that LUA objects could implement, but nobody ever does and also they shouldn't. The original stalker and anomaly do nothing with it, and all the scripts GAMMA uses do not use it. 

`net_Relcase` is meant for properly disconnecting from objects that are linked when the object gets destroyed. There isn't a use-case for this in scripting, and again is never used.

However, it still gets called and can cause a lot of work for the GC which will cause stuttering. Here is the profile of the stutter issue I was able to reproduce:

image


themrdemonized:
thanks

PR #125, opened 2025-01-22
Expose noise and luminosity values to scripting

CatspawMods:
Add def_readwrite for m_noise_progress and m_luminosity_progess, exposing them to scripting as properties of ActorMenu.get_maingame()


CatspawMods:
Canceling, misunderstood context.

PR #128, opened 2025-02-15
Adds new weapon: Magazine fed grenade launcher -> SSRS (allows modding in mag fed GLs)

Bence7661:
This PR doesn't just add a new weapon to the engine, but allows modders to create their own Mag fed GLs.
I made this code based off of the existing code for the RG6.

Currently the weapon is limited to semi-auto mode. I sadly failed to figure out how to make it work in full auto mode, so the weapon should be constricted to `fire_modes = 1` in all mods building upon this. Would love some feedback on a possible solution to this. That would make the feature complete basically.

Probably would need to move the actual shooting code out ot `FireStart()` method.
I guess it's only a "callback" when the gun starts shooting.

I also have files for a W.I.P. weapon mod for testing purposes.

### DIFF
--- a/src/xrGame/WeaponSSRS.h
+++ b/src/xrGame/WeaponSSRS.h
@@ -1,0 +1,32 @@
+#pragma once
+
+#include "rocketlauncher.h"
+#include "WeaponMagazined.h"
+#include "script_export_space.h"
+
+class CWeaponSSRS : public CRocketLauncher,
+                    public CWeaponMagazined
+{
+	typedef CRocketLauncher inheritedRL;
+	typedef CWeaponMagazined inheritedWM;
+
+public:
+	virtual ~CWeaponSSRS();
+	virtual BOOL net_Spawn(CSE_Abstract* DC);
+	virtual void Load(LPCSTR section);
+	virtual void OnEvent(NET_Packet& P, u16 type);
+
+#ifdef CROCKETLAUNCHER_CHANGE
+	virtual void UnloadRocket();
+#endif
+
+protected:
+	virtual void FireStart();
+	virtual void ReloadMagazine();
+
+DECLARE_SCRIPT_REGISTER_FUNCTION
+};
+
+add_to_type_list(CWeaponSSRS)
+#undef script_type_list
+#define script_type_list save_type_list(CWeaponSSRS)
--- a/README.md
+++ b/README.md
@@ -184,6 +184,9 @@ How to compile exes:
 6. For successful compilation, **the latest build tools with MFC and ATL libraries is required**
 
 ## Changelog
+**2025.02.07**
+* GAMEMTL_SUBITEM_COUNT increased 10 -> 20
+
 **2025.01.26**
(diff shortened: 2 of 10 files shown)


Bence7661:
The `#ifdef CROCKETLAUNCHER_CHANGE` checks could probably just be deleted since the rocket launcher will probably remain part of the repo in the future.

src/xrGame/WeaponSSRS.h
```diff
@@ -0,0 +1,32 @@
...
+                    public CWeaponMagazined
+{
+	typedef CRocketLauncher inheritedRL;
+	typedef CWeaponMagazined inheritedWM;
+
+public:
+	virtual ~CWeaponSSRS();
+	virtual BOOL net_Spawn(CSE_Abstract* DC);
+	virtual void Load(LPCSTR section);
+	virtual void OnEvent(NET_Packet& P, u16 type);
```
Bence7661, line 17: Wasn't sure if this was only code for Multiplayer modes or some other event handler, so I followed the implementation that the RG6 had.

themrdemonized:
Hi. Adding a brand new weapon class would involve some changes to script classes, particularly to class_registrator.script and global functions that involves working with clsid. Would it be possible to change existing WeaponMagazined class to support firing grenades instead of bullets? You will also have all additional functionality of that class including working fire_modes.

For now i reverted the merge commit in `wpo` branch, the `all-in-one-vs2022` branch has the merge commit

Bence7661:
> Hi. Adding a brand new weapon class would involve some changes to script classes, particularly to class_registrator.script and global functions that involves working with clsid. Would it be possible to change existing WeaponMagazined class to support firing grenades instead of bullets? You will also have all additional functionality of that class including working fire_modes.
> 
> For now i reverted the merge commit in `wpo` branch, the `all-in-one-vs2022` branch has the merge commit

I have a configured `class_registrator.script` set up with the W.I.P. Mag fed GL mod that I mentioned. But based off of your comment I'm not fully aware of the side effects that the modification would cause with other `.script` files handling clsids.

Unfortunately the GL needs to be set up in a specific way in order for it to work. The custom class I created already inherits from `CWeaponMagazined` and `CRocketLauncher`, but the `CWeaponMagazined` can't support this concept out of the box. I tried to do so, but without the custom code everything works except an actual projectile won't fire.

I think I could make this work with modifying the `CWeaponMagazined` class or the `CWeaponMagazinedWGrenade`, but I would have to rearrange/split the whole class code into 2 different modes: Normal `CWeaponMagazined/CWeaponMagazinedWGrenade` flow and the custom GL flow. Guess I could make it decide which mode it should operate in from `.ltx` files using something like this `if (pSettings->line_exist(section, "weapon_is_customgl"))`.

As I stated I don't know what `.script` files could break by adding a new clsid, but from an SRP standpoint at least it would be better to have the gun as a different class altogether. 

I prefer discussing this further on DC since it's faster/easier for me. I sent you a friend request. If you want to keep the discussion here just reject the request.

PR #129, opened 2025-02-18
Update Actor_Weapon.cpp

VodoXleb:
Changed else if to make firepos work while aiming. Not sure if it doesn't break anything
pWeapon->GetZRotatingFactor() != 1.f was preventing g_firepos from working while player is aiming


themrdemonized:
Please make it into toggle able via console variable

VodoXleb:
Added Console variable "g_firepos_zoom"

PR #134, opened 2025-03-04
Feature/ubgl alt aim separated keybinds

Bence7661:
Adds the ability to separately set keybinds for UBGL (Under Barrel Grenade Launcher) and alternat aim switch (Canted, Laser, etc).

Also adds a feature:
- The gun will "remember" what aim mode you were using before switching to the UBGL, so when you switch back it will know what aim mode to place you in.
- Added console (aimmode_remember) command to toggle this feature since the animation can get a bit weird. I can provide example video if requested. The command is also added to options menu.

Example: I have a Scope, Canted sight and UBGL on my gun. I press the alt aim keybind so now I'm aiming down the canted sight. Then I press the key to switch to the UBGL and then I either press the UBGL switch again (to put UBGL away) or press the alt aim keybind and I will end up aiming down in my canted sight again.

### DIFF
--- a/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -165,6 +165,9 @@
 	<string id="ui_mm_modded_exes_qol_crash_save_count">
 		<text>Crash Saves Count (crash_save_count)</text>
 	</string>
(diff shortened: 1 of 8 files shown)


Bence7661:
Sorry I fckd up the PR I wanted to send this to my fork and then send the fork to your repo, but I guess it's fine this way too. I can cancel and do it the way I mentioned if you prefer.

PR #137, opened 2025-03-08
Pregenerated gasmask rain noise

ZoulKrystal:
Added necessary files for it to work, crashes if gamedata\textures\shaders\gasmasks\mask_noise is missing. Should improve performance by a bit when using mask overlay with no noticeable difference (unless the texture is changed).


damoldavskiy:
Is it possible to hide it under flag? Currently modded exes don't require its own gamedata (it's safe to only install .exes, without other files), crash on missing texture is going break this feature

ZoulKrystal:
mb there, it seems it doesnt crash on missing texture, it does when its invalid (but that's with everything).
I also deleted gamedata files, I can just upload them as separate mod.

PR #139, opened 2025-03-12
Enables freelook while reloading.

Bence7661:
The title is pretty self explanatory.

I don't know why the original creator blacklisted reloading from freelook, it always felt weird for me that you can't do it.


Bence7661:
Maybe I should add console command so ppl can toggle this on and off in case it was blacklisted for a reason?

themrdemonized:
Yes, make a console command for this

PR #140, opened 2025-03-13
Script Attachments

xr-lucy:
Adds support for script attachments as well as some other goodies necessary for them to work.

Notable additions:
- Fixed hud_mode for lights
- Hidden bone transform + access to HudItemData at all times
- Small render optimization + actor shadow rework
- on_before_mouse_move callback


themrdemonized:
Thanks

PR #161, opened 2025-04-13
Indoor Gunsound Framework

Kutez:
- Make it possible to add custom gun sound for indoor, using `g_gunsnd_indoor` command to use it in game.
- To add custom sound, do `snd_shoot_indoor` and `snd_shoot_silncer_indoor` in the weapon stat.
- The command is a float from 0 to 1 to change volume of indoor and outdoor sound, for example 0.7 mean the indoor have 0.7 volume and the outdoor have 0.3 volume. 1 mean all indoor sounds, 0 mean all outdoors sounds. (0 is the default).
- Add a new parameter for HUD_SOUND_ITEM, HUD_SOUND_COLLECTION, HUD_SOUND_COLLECTION_LAYERED :PlaySound function called volume_mult allow changing the volume of sounds played, default value = 1. Require to make the command above possible.
- g_gunsnd_indoor_volume to change indoor gunsound’s volume.
More info on the mod can be found on its mod thread in Anomaly discord

### DIFF
--- a/src/xrGame/Actor_Flags.h
+++ b/src/xrGame/Actor_Flags.h
@@ -23,7 +23,6 @@ enum
 	AF_AIM_TOGGLE = (1 << 19),
 	AF_3D_PDA = (1 << 20),
 	AF_FIREPOS_ZOOM =(1 << 21),
-	AF_GUNSND_INDOOR =(1 << 22)
 };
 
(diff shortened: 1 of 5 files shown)


themrdemonized:
Thanks

PR #171, opened 2025-04-17
Lucy game object shader texture changing

xr-lucy:
- Removal of Model Pool caused a lot of unpredictable crashes, so now it's back but with a small change to work in line with the set_shader lua methods :)
- Added `get_default_shaders` and `reset_shader` to easily revert a model to its vanilla shaders without the need to store default values in a table
- Added support for getting/setting/resetting Script Attachment model shaders/textures


xr-lucy:
...don't merge yet, there's still some issue with resetting to default shader/texture, for whatever reason the default strings are nullpointers :noooo:

PR #172, opened 2025-04-20
Update v1 for Indoor Gunsound Framework

Kutez:
-  New _bak param that allow create backup indoor sounds, if there is another non _bak param, then it will automatically overwritten the _bak param. This is for my upcoming mod, which generate indoor sounds automatically.
-  New `snd_shoot_indoor_actor` for indoor actor's gunsound, it can still use the old `snd_shoot_indoor`, if it doesn't have the actor one.
- Explosion sounds for indoor: `snd_explode_indoor` (Rely on the player current position, not the nade). Add more function to Explosion.cpp to check if a sounds is there or not.


Kutez:
Also do you have the VSC plugin that doesn't make the file automatically change to UTF-8? It's really annoying for me and you.

themrdemonized:
I use visual studio 2022, it doesnt have this issue

themrdemonized:
1. Please fix the unnecessary changes to comments, its hard to see whats going with the rest of PR because of that
2. ```New _bak param that allow create backup indoor sounds, if there is another non _bak param, then it will automatically overwritten the _bak param. This is for my upcoming mod, which generate indoor sounds automatically.``` 
Is there any better way to do this? What does "backup sounds" even mean

For now your changes will reside into `kutez` branch. When you are ready with your mod i'll merge into the main branch, there were too many breaking changes lately so the updates to main will be less frequent

Kutez:
1. Sorry, so I have to use your Action to compile the exe (I don’t storage to install the compiler on my PC), so I can test it. This required me to make a commit every time, so I had to comments random things. I gonna change that if you find that annoying.
2. Yes there is. Backup sounds is an computer generated version of indoor sounds from existing one. Seemed like my mod is too hacky to be implemented on main. So I will revert it. Sorry.

themrdemonized:
1. I'll try to fix the comments this time, seems like the files in engine are mostly saved in windows-1251 encoding. Try to save to that in VSCode if you use that. Next time I would appreciate to see clean commits without that mess on your end :)
2. Maybe you can rename `_bak` to make variations of sounds such as `snd_shoot_indoor_actor1`, `snd_shoot_indoor_actor2`, etc, or similar, like how `snd_shoot` is made. Or maybe make a lua callback that will dynamically overwrite shooting sound with any sound you want, ie dynamically replace `snd_shoot` with `snd_shoot_indoor_whatever`. The second approach looks pretty customizable

Ideally the game should use already defined sound environments in SDK, it requires fixing OpenAL to use them but nobody tried to do that so far

Kutez:
2. I don't understand your first option very much. You mean just replace `_bak` part with `1/2/3'` then it will be alright to add? Why `_bak` is not ok but 1/2/3 is ok? I'm just asking.  🤔

themrdemonized:
TBH i still don't understand the intent of "backup". Why generated sounds can't be used as `snd_shoot_indoor_actor`? I assume you just generate them in separate program and you have ready sound files for indoor sounds.

Kutez:
Yea, now that I think about it I feel like a dumbass because I can just use the DLTX naming priority.

themrdemonized:
I've tried to add possibility to override current sound via Lua callback. You can set the section and line to pick new sound from the scripts

https://github.com/themrdemonized/xray-monolith/commit/47dc973b11394b7e5ad5888f2ade3468bfe44a52

What you think of this feature?

Kutez:
Very interesting. Can made my life a lot easier. Thanks.

PR #174, opened 2025-04-20
Add support for distortion on HUD geometry

nltp-ashes:
This PR adds support for distortion shaders on HUD geometry.

Most of the code is *inspired* (an exact copy) of [code](https://github.com/ixray-team/ixray-1.6-stcop/commit/801e6f0cfe890b5356cbf42a8ad8164fa8044efe) made by [Drombeys](https://github.com/Drombeys) for the IX-Ray engine. So credits goes to them for the work, I simply ported it to Anomaly's engine.


themrdemonized:
Why there are changes in dog ai files 🤣

nltp-ashes:
Ah. These are left over changes from something else I was trying to do. Well, technically, these aren't changes. Just code reformat because whoever formatted it before had serious issues.

PR #179, opened 2025-04-23
HUD <-> World Transforms

ProfLander:
Following up to my Discord ping: I made CRenderDevice authoritative over HUD <-> World transformations, applied some necessary corrections to the logic, and redirected existing implementations. Also exposed the matrix versions to Lua.

I opted not to add any extra Lua methods to CScriptRenderDevice for now, since the existing vector / matrix -based API is fine, and multiple entrypoints will just add maintenance burden down the road.

Extra details are available in the commit messages where relevant - any questions, just let me know.

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -281,6 +281,8 @@
         function setXYZi(float x, float y, float z)
         function setXYZi(Fvector)
         Fvector getHPB()
+        function hud_to_world()
+        function world_to_hud()
     }
 
     class particle_object {
@@ -529,7 +531,7 @@
         function GetZoomRotateTime()
         function SetZoomRotateTime(float)
     }
-
+	
     enum rq_target { (sum them up to target multiple types)
         rqtNone     = 0,
         rqtObject   = 1,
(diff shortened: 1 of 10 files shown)


themrdemonized:
`Use transform instead of transform_tiny`
This gives me wrong results when using in my upcoming Weapon Cover Tilt update, hence the reason i did own exports in the first place

When i get the bone position vector of fire_bone of a gun and apply `hud_to_world` to a vector, with transform it looks misplaced

AnomalyDX11AVX 2025-04-24 17-23-52 After

While previously with transform_tiny it looked alright

AnomalyDX11AVX 2025-04-24 17-30-48 After

```lua
local fire_bone = SYS_GetParam(0, hud_sec, "fire_bone", "wpn_body")
local offset = SYS_GetParam(0, hud_sec, "fire_point", "0,0,0")
if isSilencedWeapon(wpn) then
fire_bone = SYS_GetParam(0, hud_sec, "fire_bone_silencer", fire_bone)
offset = SYS_GetParam(0, hud_sec, "fire_point_silencer", offset)
end
offset = utils_data.string_to_vector(offset)
local pos = wpn:bone_position(fire_bone, true)
local mat = wpn:bone_transform(fire_bone, true)
mat:transform_tiny(pos, offset)
pos:hud_to_world()
```

PR #181, opened 2025-04-23
Fix inconsistent .def syntax in script_render_device_script.cpp

ProfLander:
A tiny bit of fix errata that fell out when I shook the 3D shooting branch.

Appears to be a (benign?) syntax error, though VS can't confirm as it lives inside a macro.

Disregard if this is intentional, and simply me not understanding luabind well enough yet.


themrdemonized:
The way its written originally is `app_ready()` function resides in global namespace, so its called like that. Your PR moves it into `device()` namespace, so changed to `device().app_ready()`. While it might be an oversight on OG devs side, its a valid code. The vanilla scripts don't use that, and idk what mods use it. Nevertheless for compatibility reasons its better leave as is imo

ProfLander:
Interesting, thanks for the breakdown - definitely sounds better left be, so I'll close this out and nuke it from my local branches 👍

PR #183, opened 2025-04-24
Shader-Based Crosshair

ProfLander:
Replaces CHUDCrosshair with an upgraded version that supports 3D projection, wireframe and shader modes, and controllable depth sizing.
New options are exposed via globals, with entries in the settings menu for mode and size, and texture / shader string console commands for user and modder customization.

Should be 100% compatible save for the DLTX patch, which introduces a ~1px visual difference for the sake of a nicer scaling visual in wireframe mode.


ProfLander:
Checked myself on the weighty claim of 100% accuracy: Dispersion was in fact having an effect, but was (wilfully?) minimized into irrelevance via default LTX values.

Have plugged it in as a near-size offset, scaled by the far plane to bring it into a reasonable coordinate space, allowing it to be hidden until the user raises the respective setting above 1. This seems mildly more accurate to weapon spread in practice versus the original impl; I don't think either is mathematically watertight, but it should suffice to fool the eye for `cl_dynamic_crosshair` users.

I also double-checked the point about the 1px visual change versus via GIMP, and got the same 3x3px cross as vanilla under default settings, so I'm pretty happy with this compatibility-wise now.

PR #190, opened 2025-04-27
Implement `g_firepos`

ProfLander:
Rebased and re-merged onto mainline, as requested!

_Original Description_
---
Here it is, the big kahuna. A full-treatment reimplementation of g_firepos and g_firepos_zoom, making sure to respect vanilla behaviour, while extending the picking code into a fully-fledged camera + arms system, both in ways hinted at in leftover (now removed) comments, and beyond to further abstraction.

Rolls up the rest of my PRs on account of hard dependencies, as mentioned.
Specific details are available in each commit message; any questions or amends, let me know.

In the meantime, sleep 🥴


ProfLander:
Fixed some subtle logical errors / omissions in course of sorting out the crosshair sizing; various aspects of crosshair positioning and nearwall are now properly correct.

As a consequence I've had to disable dispersion visualization until it can be given a proper treatment, as the numbers going into the crosshair don't start at zero and scale up, so create chaos in its delicate sizing operations.

Not a terrible loss for now imo, as it wasn't visible without editing the crosshair's LTX parameters, but I'll table it for a fix later down the line so the feature can at least be given a proper treatment.

PR #196, opened 2025-04-30
Fix MAS crash, correct handling of scope near-wall

ProfLander:
`player_hud::OnFrame` was erroneously trying to call trace updates on the scope attachment slot, which appears to be of some simpler type that lacks the requisite strata to call said functions without crashing.

This was a paranoia measure for robustness anyway, so has been removed, and the transform handler updated to properly position MAS scopes during near-wall.


themrdemonized:
Contained in https://github.com/themrdemonized/xray-monolith/pull/197,

PR #197, opened 2025-04-30
Crosshair: Far size option

ProfLander:
Mumuskeh pointed out the absence of non-LTX far plane scaling for the crosshair, so I've added `g_crosshair_far_size` to control its size at the far plane.


themrdemonized:
Contains https://github.com/themrdemonized/xray-monolith/pull/196

PR #198, opened 2025-04-30
Lua trace API improvements

ProfLander:
Adds a new `TT_ACTOR` variant to `ETraceType` (at value -1, both to avoid breaking the existing values, and as a semantic indication of default-ness,) and modifies the no-argument implementations of `g_get_target_*` to use it instead of `TT_CAMERA`.

This allows users of the Lua API to fetch what the engine currently considers the 'active' trace (i.e. the camera with `g_nearwall_trace 0`, and weapon / device / camera with `g_nearwall 1`) where previously it would default to the camera, and so only behave intuitively under vanilla settings.

Should prevent some reverse-engineer-and-reimplement headaches on the part of API consumers.

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -175,8 +175,10 @@
         box.size = vector():set(0.5,1,0.5)
 
 
-    // Query data from actor ray queries; no parameter defaults to camera ray
+    // Query data from actor ray queries
+    // Actor = Pick active ray from Weapon / Device / Camera
 	enum ETraceTarget {
+		Actor       = -1,
 		Camera      = 0,
 		Weapon      = 1,
 		Device      = 2,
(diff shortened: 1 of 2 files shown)


themrdemonized:
I think the default setting should still be camera. I'm afraid changing that will lead to discrepancies in scripts if used with `g_firepos 1` or when a 3rd party script toggles this cvar. If you want to call explicitly for actor, you should specify it in function parameter in your scripts.

themrdemonized:
Or make a console variable for default ray pick option

ProfLander:
That's fair - whether the original API is used with a semantic of "look / interact target" or "shoot target" is naturally ambiguous now those notions have been separated, so erring on the side of caution is sensible.

My main concern is making sure modders don't have to mess around reading cvars to figure out which trace they should use, so the addition of `TT_ACTOR` is good enough.

I've opened #200 to reinstate `TT_CAMERA` as the default.

PR #201, opened 2025-05-01
Scriptable Nearwall

ProfLander:
Adds `actor_hud_on_nearwall` for intercepting and modifying the near-wall transforms from Lua.

It receives transform index, offset distance / direction, mutable offset matrix, and a mutable override flag for suppressing built-in behaviour. The barrel trace is updated after the offset is applied, so the crosshair and bullet impact point both respect it.

This should be sufficient to hook WCT's per-weapon offsetting logic into the new system without too much trouble. For demo purposes, I've uploaded my [basic testing script](https://gist.github.com/ProfLander/67ee5f8e4c49334a3f946accd87fb696) as a gist. Drop it into gamedata, and weapons + devices should start tilting up instead of moving backward when approaching a wall with `g_nearwall 2`.

### DIFF
--- a/gamedata/scripts/callbacks_gameobject.script
+++ b/gamedata/scripts/callbacks_gameobject.script
@@ -134,6 +134,28 @@ end
 -- 	end)
 -- end
 
+-- HUD item callbacks
+
+-- Near-wall transform callback
+-- Intercepts the transforms applied to the HUD when near a wall
+-- Receives the following table:
+-- {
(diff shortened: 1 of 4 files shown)


ProfLander:
Had to rebase this back onto `lander` and apply a small patch to prevent compile error following the addition of `SCOPE_ATTACH_IDX` offset application in #196.

PR #210, opened 2025-05-03
Crosshair Sizing Fixes Again

ProfLander:
Crosshair sizing was inaccurate to vanilla, so has been switched to a pixel-based metric to ensure proper 3x3 display independent of resolution and FOV. New commands have been added to control near/far scaling range, and the default distance metrics have been modified accordingly.

It was also incorrectly applying crosshair roll when `g_firepos` was turned off, which is now gated by a conditional.


themrdemonized:
Please make changes to the current PR related to the problem instead of creating new PR, it will ease the life for both of us. Mark the PR as draft if its not ready to be merged

ProfLander:
Noted - is that possible with a PR that's already been merged? I assumed not, but if so then I'll target the base branch if the suboptimal situation of after-the-fact bugs comes up in future.

And apologies for the last-minute epsilon amend; that was a shower thought that occurred while away from the machine after the fact.

themrdemonized:
Make a new PR and that will be your new changes base

PR #213, opened 2025-05-04
Fix scriptable near-wall jitter

ProfLander:
Desync between the `OnFrame` near-wall application and `UpdateCL` near-wall update was causing the post-script trace transform to leak into subsequent frames, thus causing jitter whenever the modified aim point disagreed with the ground-truth weapon / item traces.

This moves said update logic into `OnFrame`, which corrects the order-of-execution problem and results in smooth offsetting.

### DIFF
--- a/src/xrGame/HudItem.cpp
+++ b/src/xrGame/HudItem.cpp
@@ -515,7 +515,7 @@ float CHudItem::GetTargetNearWallOffset()
 	return 0.f;
 }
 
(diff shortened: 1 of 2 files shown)


I-Am-Everyday-People:
Could you add a condition to minimise or exclude the weapon movement when game.actor_lower_weapon(true) is present? Using weapon lowered sprint the gun rises awkwardly when running through bushes , tall grasses etc.

PR #215, opened 2025-05-05
Improved categorization and definition machinery for settings menu

ProfLander:
- Implemented utility functions to reduce definition boilerplate
- Vanilla-oriented categories for easier high-level navigation
- Feature-oriented subcategories for better discoverability and less scrolling
- Fixed misplaced crosshair distance animation setting, missing translation string


ProfLander:
Have tested each of the new UI control abstractions with a range of applicable settings, and all seems to be working as intended.

ProfLander:
Came up with some more ways to cut out boilerplate and improve UX; will un-draft again once finished.

ProfLander:
That should about cover it; option definition is now backed by a syntactically-consistent EDSL with little to no loss in expressive power versus raw option tables, organized into a series of small self-documenting modular scripts to separate data from implementation, and papers over all existing repetition cases (ex. `type` and `val` definition, `cmd = id`, and so on.)

I also noted that the HDR live edit interface broke during the move, due to hardcoding against the original option table structure; I've rewritten the relevant parts to encapsulate option introspection within the HDR10 module, which works out nicely as a self-contained example of custom UI.

ProfLander:
`settings-menu-refactor.final.2.bak.final.very-final` 🥴

ProfLander:
Also, this is liable to collide with #214 and possibly other upcoming PRs that add new options to the menu.
Let me know how you'd like to handle it; I'm happy to do merges / rebases this end if it's more convenient. Will smash everything together on a local `lander` branch in the meantime.

themrdemonized:
Merge changes from that PR into this and I close that PR

ProfLander:
Almost there - fixed a last-minute showstopper with reticle interpolation error checking, just need to amend one of the translation strings w.r.t. the new layout.

themrdemonized:
I decided to merge that branch into `lander`, so that the old base won't be wasted. You should resolve conflicts now

ProfLander:
Done and done 👍

themrdemonized:
I've added new function to get cvar bounds from the engine and integrated into "track" element builder, in case you need to update cvars you don't need to change min max in the script files

ProfLander:
Great idea, that boils down the definition surface even further!

PR #224, opened 2025-05-11
Multiple crosshairs

ProfLander:
Rewrites and restructures crosshair handling to allow display of more than one at a time, and gives each its own independent settings.

- Crosshairs are grouped into Camera, Weapon, Device
- Each group has a Near and Far crosshair
  - Near crosshair is positioned by tracing into the world
  - Far crosshair is always drawn at the far plane
- Near crosshairs scale from their set size to that of their corresponding Far crosshair based on distance
- Distance / recon indicators now respect occlusion fade, and have an opacity limit to recapture vanilla behaviour
- Each group may draw a multi-segment line from the barrel through its Near and Far crosshairs
- The existing cross texture has been replaced with a trio of overlayable graphics
- Under default settings, vanilla behaviour is recaptured via the Camera Far crosshair
- Console commands have been implemented appropriately, and exposed in the corresponding options page

I've also included some misc fix commits for related systems that showed up in the course of building this feature; those can be moved to a separate PR if desired.


ProfLander:
Marking as draft, as I need to fix an issue with the distance interpolation setting; under some condition, `NaN` ends up in its output during map load, and results in non-working crosshairs until the setting is toggled off and on again.

ProfLander:
That should about do it; interpolation fix, per-crosshair coloring, and a small orientation fix for the knife crosshair.

Recon parameters (how fast names show up, etc.) could also be broken out into per-crosshair settings, but I'll save that for future work, since this is already getting quite large.

ProfLander:
Drafting again after discovering a showstopper; moving the weapon trace into foliage causes massive framerate drops. Will investigate.

ProfLander:
Showstopper fixed, plus a pass over third-person to ensure proper operation.

Bit of a delay on account of trying to fix weapon crosshair display in the 'freelook third person' camera mode, as pointing the camera backward along the barrel causes the projection to flip.

No dice with that so far, but it's minor given that shooting isn't yet practical in that mode, so have undrafted and will keep investigating.

PR #233, opened 2025-05-23
Export CurrentGameUI()->m_pMessagesWnd to Lua

nltp-ashes:
This PR exports `CurrentGameUI()->m_pMessagesWnd` to Lua under the name `ActorMenu.get_messages_menu()`.

It should be noted that right now, this is exported as a `CUIWindow`, instead of it's real type `CUIMessagesWindow`. This means that anyone trying to export additional functions from it will not work.

I do not have the skills to export it as `CUIMessagesWindow`, so I am leaving it as `CUIWindow`, as it is sufficient for my needs.

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -244,6 +244,10 @@
         //    level.add_bullet(t)
     }
 
+    namespace ActorMenu {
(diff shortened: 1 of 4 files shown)


themrdemonized:
Fixed: Exported as `CUIMessagesWindow*` now

PR #237, opened 2025-05-28
Fix popup positioning for tall popups in tasks/map tab of PDA.

lulnope:
Tasks/Map tab popups often got clipped by PDA window. This PR fixes that for all cases where popup is less tall than PDA window. If poup is taller than PDA window it will get clipped at the top.

### DIFF
--- a/src/xrGame/ui/UIMapWnd.cpp
+++ b/src/xrGame/ui/UIMapWnd.cpp
@@ -927,25 +927,16 @@ void CUIMapWnd::ShowHintTask(CGameTask* task, CUIWindow* owner)
 	{
 		m_map_location_hint->SetInfoTask(task);
 		m_map_location_hint->SetOwner(owner);
-		ShowHint(true);
+		ShowHint();
 		return;
 	}
 	HideCurHint();
 }
 
-void CUIMapWnd::ShowHint(bool extra)
+void CUIMapWnd::ShowHint()
 {
-	Frect vis_rect;
-	if (extra)
-	{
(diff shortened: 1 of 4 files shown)


themrdemonized:
Can you show a screenshot of before and after the fix?

Bence7661:
> Can you show a screenshot of before and after the fix?

image
image

StalkerBoss:
Still not fixed when TAB is pressed.
AnomalyDX11AVX Screenshot 2025 05 31 - 22 21 44 54

lulnope:
@StalkerBoss 
Tell me more. It's my first time seeing the functionality. In my GAMMA installation pressing TAB opens inventory. How do I access the functionality you screenshotted? Is it the "Task Info" keybind? Task Info for me only shows the task title, not the task description, can i configure it to work like on your screenshot?

Anyway this seems to pertain to a different functionality so it probably shouldn't be discussed here but correct me if that's an incorrect statement.

StalkerBoss:
> @StalkerBoss Tell me more. It's my first time seeing the functionality. In my GAMMA installation pressing TAB opens inventory. How do I access the functionality you screenshotted? Is it the "Task Info" keybind? Task Info for me only shows the task title, not the task description, can i configure it to work like on your screenshot?
> 
> Anyway this seems to pertain to a different functionality so it probably shouldn't be discussed here but correct me if that's an incorrect statement.

Its absolutely base game feature, you can check it even in Vanilla:
Bosspack Screenshot 2025 06 01 - 16 05 13 53
Idk why it`s cut from gamma, not gamma player.

PR #240, opened 2025-06-01
Anomaly Scripting 2.0

ProfLander:
# Table of Contents

1.  [Clean Lua Reintegration](#orgd07fdb6)
2.  [Existing `CScriptEngine` and `CScriptStorage` code migrated to Lua](#org44cc97f)
3.  [Script Compilers And Macros (S.C.A.M.)](#org8e5bd26)
4.  [Improved supporting code](#orgc6b12c7)
    1.  [`print`](#org026f22e)
    2.  [`require`](#orgd8c82b9)
    3.  [`import`](#org6f8c146)
    4.  [`prelude`](#org1e7fed6)
    5.  [`axr_main`](#orgd362f61)
    6.  [`options`](#org7e4d54f)
    7.  [`patches`](#org65f6771)
    8.  [`dxml`](#org9a9b558)
    9.  [`moved`](#org113db7c)
5.  [Proposed Best Practices](#org5e5b2b0)
    1.  [Baseline](#orge1ea7cc)
    2.  [Core Language](#org01630ec)
    3.  [Program Structure](#orgad4624e)
6.  [Frequently Asked Questions and Common Statements](#orgfc2da53)
    1.  [Are there any fundamental underlying differences between Stalker Lua and normal human Lua?](#org3994b63)
    2.  [Will I have to rewrite all my scripts to work with Scripting 2.0?](#org21436b3)
    3.  [Will my text editor / IDE be able to highlight and hint new-style scripts?](#orgac305be)
    4.  [What about a Lua package manager, like Luarocks?](#org16f3e76)
    5.  [Why choose `.lua` modules over Anomaly's `.script`?](#orgb28edbd)
    6.  [I just wanted folders, none of this init.lua, boot, macro, or Fennel stuff](#org2a8eec1)
    7.  [It may be optional to modders, but you're forcing it on contributors to modded exes, and people reading its code as reference](#org2fcc0f1)
    8.  [What if I don't want to learn Fennel to be able to mod the new script engine?](#org39a3a8e)
    9.  [Won't multiple languages fragment the scripting ecosystem?](#orgd75ac5d)
    10. [Not specifying extension in `require` or `import` is weird](#orga9e84db)
    11. [Being able to import JSON and other data files as first-class scripts is weird](#orgaeccce9)
7.  [Motivations](#org1b8bb80)
    1.  [Selfish](#org78a36cf)
    2.  [Altruistic](#org110b126)
8.  [Closing Notes](#org068aef4)

A next-generation scripting environment for S.T.A.L.K.E.R: Anomaly, aiming to embrace the full power of idiomatic Lua without breaking vanilla compatibility.

<a id="orgd07fdb6"></a>

# Clean Lua Reintegration

By default, Anomaly's Lua integration is heavily specialized around its extensions to the language. Namely: Auto-loading scripts when the corresponding global is accessed, evaluating each script within a sandbox environment which inherits from `_G`, and injecting that environment as a named field of `_G` once script execution is complete.

This provides a simple and friendly interface for the everyman stalker, but prevents the script engine (and by extension, modders) from leveraging the full suite of Lua language features.

For instance:

-   'Pure Lua' libraries (such as Anomaly's own variant of SLAXML) cannot be used without first adapting them into a `.script` and patching up any violated assumptions
    -   A functioning `print` primitive for runtime status reporting
    -   Ready access to loaded modules via `package.loaded`
    -   A `require` primitive that addresses the script root (`<anomaly>/gamedata/scripts`) rather than the binary root (`<anomaly>/bin`)
    -   Access to submodules stored in nested directories, which Anomaly's present flat structure does not support

To solve this, it is necessary to hoist Anomaly's extended semantics into an optional quantity - applicable only to `.script` files - and configure the Lua environment such that raw `.lua` files are functional, and treated as first-class members of the ecosystem.

<a id="org44cc97f"></a>

# Existing `CScriptEngine` and `CScriptStorage` code migrated to Lua

Relevant members of `CScriptStorage` (previously an organizational superclass of `CScriptEngine`) have been moved into `CScriptEngine` itself in the interest of transparency.

`CScriptStorage` is now a free-standing value store for caching persistent data across instances of the script engine, and is currently used to memoize I/O misses and reduce runtime script auto-load / reload overhead.

`CScriptEngine` itself now provides a minimal engine <-> Lua bridge surface, and passes control to `gamedata/scripts/init.lua` to perform the boot process proper.

New control flow is as follows:

-   `CScriptEngine`
    -   Configure Lua superglobals to interface with engine-side dependencies
        -   `_OBJECT_FACTORY`
            -   Engine-side object factory, passed to class registrators
        -   `_SCRIPT_STORAGE`
            -   The aforementioned free-standing value store
        -   `_LOAD_FILE`
            -   Low-level engine-side loader working around `IReader::r_stringZ`'s inability to consistently load script files across loos-file and `.db` storage
    -   Load `init.lua` from the X-Ray FS and execute it with an appropriate error handler

-   -> `init.lua`
    Establishes the minimal environment necessary to hand off the boot process to submodules.
    -   Setup minimal `package.loaders` machinery for `require` to work
        -   Implicitly registers `lua` file extension search paths
    -   Hand off control to the `boot` module

-   -> `boot/init.lua`
    Establishes a functioning Lua environment with respect to the engine
    -   Disable dangerous built-in globals
    -   Emplace working `print` function
    -   Configure `package.paths` machinery
    -   Configure robust `package.loaders` machinery
    -   Configure Engine -> Lua lookup interface
    -   Return control to `init.lua`

-   <- `init.lua`
    -   Hand control to the `amx` module

-   -> `amx/init.lua`
    Anomaly Modded Exes entrypoint; establishes a fully-featured Lua environment capable of hosting Anomaly scripting
    -   Setup the `SCAM` (Script Compilers And Macros) environment
        -   Clears `package.path` state and configures extensible script compilation machinery
    -   Introduce Anomaly's original scripting semantics to the `SCAM` compiler
    -   Register auto-patching machinery via `patches` module
    -   Load `_G` (a.k.a. `_g.script`)
        -   Anomaly's original script entrypoint, now hoisted into the extended semantics via `SCAM`
    -   Register script classes via `classes` module
    -   Run common scripts via `scripts` module
        -   Defined in `configs/script.ltx`
        -   `dxml`
            -   Gathers and applies DXML scripts
        -   `options`
            -   Configures vanilla options menu

-   <- <- `CScriptEngine`
    -   Lua environment is now fully configured, and ready the invocation of `axr_main`'s `on_game_start` callback

<a id="org8e5bd26"></a>

# Script Compilers And Macros (S.C.A.M.)

SCAM is a 'Language-Oriented Programming' environment for Lua, written specifically for Scripting 2.0.

It provides extensible machinery for processing textual source into runnable Lua code, is named in the Bandit tradition of languages like Scheme and Racket, and derives its basic interface from the latter:

Scripts may now specify a `=*= lang: path/to/macro.function =*=` header on their first line instructing the engine to pass their source through a `String -> Function` transformer defined in some other module, which takes source text, and produces runnable Lua code.

The header line may be commented for compatibility with language servers and syntax highlighters. Default macro association may also be discharged by file extension, with the header acting as means to override.

This allows a given script's syntax and semantics to by manipulated arbitrarily by another script.

The foundational example is Anomaly's existing backend metatable manipulation -  trying to auto-load scripts when the corresponding global is accessed, and storing the results into `_G` as script-named fields. This behaviour has been refactored into a metatable interface over Lua's standard `package.loaded` table, codified as the `wua` language (Lua for Anomawy; Wua) and associated with the `.script` extension.

Further examples:

-   Lua -> Lua language compilers
    Means to express logic in other languages, while retaining direct access to the Lua environment, and never leaving the pure LuaJIT ecosystem.
    -   `scam/lua`
        -   Simple `loadstring` wrapper allowing inherited execution environments and detailed error reporting
        -   Invoked automatically for files with the `.lua` extension
    
    -   `scam/fennel`
        -   An alternative Lisp syntax with Lua semantics
        -   Provided by `packages/lib/fennel`
        -   Invoked automatically for files with the `.fnl` extension
        -   Provides its own extensive compile-time macro system, whose entrypoint is available via `scam/fennel/macro.compile`
            -   Modules loaded by this entrypoint are available both to Fennel's internal macro machinery, and via `package.loaded` so macros implemented in other languages can apply them directly to parsed AST
    
    -   `scam/teal`
        -   Statically-typed variant of Lua
        -   Provided by `packages/lib/tl`
    
    -   Other possibilities:
        -   MoonScript
            -   Imperative language aiming to make Lua more ergonomic
        -   Urn
            -   Full-fledged Lisp compiler with full Lisp semantics
                -   Not for the faint of heart

-   Backend syntax manipulation
    -   Script unlocalization
        -   The engine-side unlocalization data model has been reimplemented as `amx/unlocalize`
        -   For compatibility, the `.script` unlocalization system has been reimplemented as a source string manipulation inside `amx/wua/unlocalize`
        -   Fennel modules are unlocalized using the native Fennel macros in `scam/fennel/unlocalize`
            -   This can be considered an 'ideal case' use of language macros, as it manipulates Fennel AST directly instead of matching and modifying source text line-by-line
                -   Such manipulations are also possible against Lua AST, given an appropriate parsing and pretty-printing library, but introduce concerns around line number metadata that are otherwise papered over by the compiler in Fennel's case
            -   As a result of manipulating AST, it is forced to operate within the standards of the target language and implement a categorically-sound 'call continuation' pattern within it to export private data
                -   This guarantees a syntactically and semantically sound program after transformation

-   Extension languages, which restrict and / or extend the syntax and semantics of other language macros
    -   Domain-specific languages
        -   Allows the full computational power of Lua to be retained within a DSL, but its use to be restricted to the domain in question to eliminate broad classes of user-error bug
        -   For example, an `options` language with its environment access limited to composable UI control constructors, and output via module return value
            -   Still able to discharge log using full Lua language features, but definitionally incapable of modifying any game state beyond options menu
    -   Feature extension languages
        -   Layer advanced language features like static typing, dependent types, linear types, etc. on top of existing implementations

-   'Plain Old Data' languages
    -   JSON, YAML, TOML, XML, INI, potentially LTX, etc
    -   Would allow such files to be `import` -ed directly into their corresponding Lua table representation as modules, thus reducing boilerplate and obviating the relationship between code and data

<a id="orgc6b12c7"></a>

# Improved supporting code

<a id="org026f22e"></a>

## `print`

Now functions as expected; defers to engine machinery in the same manner as `printf`, but with Lua's standard variadic syntax instead of an explicit format string.

<a id="orgd8c82b9"></a>

## `require`

Now behaves intuitively, taking an absolute extensionless path rooted in `gamedata/scripts`, and allowing import of any registered file extension.

<a id="org6f8c146"></a>

## `import`

Expanded `require` with unix path support. Implemented as its own set of functions to avoid interfering with Lua's idiomatic absolute-relative-to-project-root semantic.

-   Paths are relative by default, can be prefixed with `/` to search from the root
-   `.` / `..` / `...` / etc. refer to the parent directory
-   `*` segments import all modules in the given directory
    -   Useful for automatic runtime module import
        -   Allows direct submodules to discharge their load-time behaviour
            -   ex. Perform further imports in a structured and ordered manner
    -   Can also be used to partial match against names, such as `modxml_*`
-   `**` segments recursively imports all modules in the given directory
    -   Useful for permissive automatic runtime module import, when per-submodule ordering is unnecessary
        -   ex. Loading free-standing scripts within a nested directory tree, without needing `init` modules to manage individual subtrees

Three implementations are provided to support different use-cases:

`import` yields one or more module values in the form of lua multi-values, which can drive the input of variadic functions or be collected into a table.

`import_list` yields a table containing module values.

`import_table` yields a module name -> module value map.

<a id="org1e7fed6"></a>

## `prelude`

Standard library for Anomaly modding. For implementing common computations available to all scripts, and reducing the sum total of boilerplate code modders have to write.

<a id="orgd362f61"></a>

## `axr_main`

-   Now imports the contents of `gamedata/scripts` recursively to allow `on_game_start` entrypoints within nested modules

<a id="org7e4d54f"></a>

## `options`

-   Takes advantage of Lua's module infrastructure to further reduce definition surface
-   Removed `ui_options_modded_exes` in favor of `options` module
-   `options` submodules are loaded automatically in FS order, allowing the addition of categories beyond the existing Modded Exes
    -   Modded Exes is stored in `options/_modded_exes` accordingly, to ensure it remains topmost if new pages are added
-   Data tables representing controls are now returned directly from each page's script, and pop out of `require`, allowing them to be composed directly into their parent module's data table
-   Reimplemented all but HDR10 in Fennel, as its homoiconic syntax and last-expression-is-return-value semantic are well-suited to a declarative DSL

<a id="org65f6771"></a>

## `patches`

-   When a given script `S` is loaded, the script engine will try to load the corresponding `patches/S` module
-   Allows a degree of 'order independent' patching; the existing FS-ordered `aaaa_load_me_first` and `zzzz_load_me_last` pattern now only need be applied within the scope of individual patch targets, rather than globally.
-   Works with nested directories, so loading `foo/bar/baz` will load `patches/foo/bar/baz` directly after
-   Existing root patches have been refactored into submodules of `patches`

<a id="org9a9b558"></a>

## `dxml`

-   SLAXML has been moved into `packages/slaxml` as an unmodified vanilla distribution from GitHub; partly to simplify upgrades in the unlikely case that it sees further releases, and partly for the sake of transparency
-   Anomaly-specific modifications have been moved into `patches/slaxml` as auto-loaded monkey patches
-   Now loads both the original `modxml_*` specification (plus new file extensions, and support for modules in nested directories) as well as all modules in the `dxml` directory
    -   Allows DXML scripts to be organized within mods' individual folders, or within the `dxml` root

<a id="org113db7c"></a>

## `moved`

Automatic path remapping, for moving old `.script` files out of the flat root structure and into nested directories while maintaining compability.

-   Implemented as a map from old script name to a table of parameters:
    -   `to` defines the new module path
    -   `if_not_overwritten` prevents the redirect if a copy of the original file still exists
        -   Intended for cases like `scopeRadii` -> `2d_scopes/scope_radii`, where the original file is intended to be modified and distributed with downstream mods as an overwrite
            -   Allows the new nested version to act as default, while still allowing replacements to overwrite it
-   Currently populated with all pre-2.0 modded exes scripts which could reasonably be expected to expose members

<a id="org5e5b2b0"></a>

# Proposed Best Practices

Given that these changes entail a significant sea-change in the Anomaly scripting ecosystem, I've made an effort to design the new modder-facing features - such as the use of `patches` and `dxml` as container directories for their respective input scripts - in a manner that will encourage proper organization of modder code, and by extension, users' final MO2 directories.

While in practice this is liable to end up in the familiar free-and-wild state that Stalker modding is reputed for - with some users doing things by the book and others implementing their vision by any means - I feel it wise to lay out a standard to which mods might adhere, in the hope that, one day, Anomaly mod directories might be as clean and well-structured as its best-in-class contemporaries.

As such, this is very much open to feedback and change, as well as potential addition of new automated systems before merge if the effort:payoff ratio justifies it at this significant juncture.

To wit:

<a id="orge1ea7cc"></a>

## Baseline

The foundation of Scripting 2.0 is standard Lua. Ergo, the [Lua 5.1 Reference Manual](https://www.lua.org/manual/5.1/manual.html), as this is the version targeted by Anomaly's LuaJIT subsystem.

The concepts laid out therein should (barring local corner-cases like sandboxing away program exit and arbitrary IO writes) be directly applicable to Lua written for Anomaly, as should the wealth of pedagogic materials available for the language.

This implicitly means standardizing with all other extant Lua 5.1 code, which is a significant win in terms of language usability and incumbent domain knowledge.

<a id="org01630ec"></a>

## Core Language

As such, I propose that `.lua` become the recommended language for new mods going forward, potentially with the addition of a new sub-language offering a clean reimplementation of `.script`'s simplifying features without straying quite as far from cleanly-namespaced idiomatic Lua.

`.script` is pleasant to use, but not ideal on account of needing to emulate Anomaly's original (partially-unconfigured) Lua environment, and so should be considered a legacy artefact. It is a vital component, as all of vanilla is written in it, but is a subpar tool for new code on account of its existing limitations.

Alternately, Teal could be posed as a new standard, as it represents Lua with the addition of strong typing. This would result in better programs overall, but increase the difficulty floor for scripting in Anomaly, which is likely undesirable.

<a id="orgad4624e"></a>

## Program Structure

Moving to structure, populating `gamedata/scripts` with loose files should be discouraged in the face of nestable directories, as each mod can now have its own, and keep it as it pleases. As such, the root should be seen as the domain of the engine, and reserved for `init.lua`.

The top-level extensible module pattern - used for `patches` and `dxml` - should be encouraged for mods with a remit of broad extensibility, as it establishes a single source of truth for features which need to invoke an arbitrary set of sub-modules at runtime in order to gather final input data, and presents a better-formalized approach to domain-specific extensions than raw monkey-patching.

<a id="orgfc2da53"></a>

# Frequently Asked Questions and Common Statements

<a id="org3994b63"></a>

## Are there any fundamental underlying differences between Stalker Lua and normal human Lua?

Civilian Lua doesn't auto-load scripts when you refer to a global that doesn't exist yet, doesn't automatically sandbox loaded scripts' globals into their environment table, and doesn't load the results into global scope.

Anomaly Lua also has an unconfigured `package.loaders` and `package.path` by default (though these were previously only relevant to niche invocations of `require`,) rooting in the `bin` directory instead of `gamedata/scripts`

<a id="org21436b3"></a>

## Will I have to rewrite all my scripts to work with Scripting 2.0?

No, `.script` files stored in `gamedata/scripts` will continue to work as they always did.

New-style nested modules can be accessed from old-style scripts via `_G["path/to/module"]`.

`path.to.module` was considered as an option, but introduces unsoundness by allowing child modules to overwrite parent fields, so has been avoided.

(This may sound like a nice feature, but results in action-at-distance at the module / filesystem level, which is best avoided to guarantee sound program structure. Such overwrites should instead be implemented as monkey-patches in the `patches` module tree, which provides an explicit and robust framework for such cases.)

Existing mods are recommended to migrate to the new features, and new ones are recommended to use them instead of `.script`, but it is not mandatory.

<a id="orgac305be"></a>

## Will my text editor / IDE be able to highlight and hint new-style scripts?

Yes! Provided that it adheres to the now-standard LSP (Language Server Protocol) standard, which offers a One Plugin -> Any Language static analyzer compatible with most extensible editors.

<a id="org16f3e76"></a>

## What about a Lua package manager, like Luarocks?

Theoretically, yes. Luarocks could be configured to populate a root inside of `gamedata/scripts`, which could be added to `packages.loaded` to hook it into require.

In practice, Luarocks is a terrible mess full of out-of-date packages, which doesn't stratify rigidly enough to prevent package maintainers from shipping broken makefiles and configuration typos.

As such, replacing `fennel` and `slaxml` with their rocks distributions ended up adding more complexity than it eliminated, so in-line distribution was chosen for flexibility.

<a id="orgb28edbd"></a>

## Why choose `.lua` modules over Anomaly's `.script`?

`.script` sacrifices computational power and expressibility in the name of comfort, `.lua` gives you the full capabilities of the language.

<a id="org2a8eec1"></a>

## I just wanted folders, none of this init.lua, boot, macro, or Fennel stuff

The new Lua-side script engine initialization machinery is a necessary part of implementing these features, as to do so in special-cased C++ would worsen Anomaly's existing engine-side maintainability problem, while adding only a small modicum of value in trade.

Doing a complete and correct reintegration in pure Lua lessens the maintainability issue, opens up new classes of potential mod, and makes it easier to amend any potential design problems that either still exist, or were (hopefully not) introduced during migration.

As for `init.lua` in a more general sense, it is an implicit necessity of nestable directories, as each one must be able to define a module to represent itself in order to fully satisfy the expectations of standard Lua.

The top-level instance serves dual purpose: First as entrypoint, and second to canonize Anomaly's `gamedata/scripts` folder as a self-contained valid Lua program.

<a id="org2fcc0f1"></a>

## It may be optional to modders, but you're forcing it on contributors to modded exes, and people reading its code as reference

The new `gamedata/scripts` structure, as well as the more advanced new features like extensible languages, are a corrolary (and worked example) of the power offered to Anomaly by a properly-integrated standard Lua environment.

Vanilla is already an abundant reference for `.script` files, and use of the new features in their parent codebase both allows modded exes' scripts to benefit, and provides equally-important reference material for the features themselves.

<a id="org39a3a8e"></a>

## What if I don't want to learn Fennel to be able to mod the new script engine?

What if someone doesn't want to learn Lua to be able to mod Anomaly? Or C++ to be able to mod major engine components? Anomaly's entrenched languages have the benefit of tenure, but not of being the right tool for any given job.

Fennel was chosen for applicable components (and selective rewrites) due to its status as a functional-paradigm expression language; this eliminates broad categories of programmer-error bugs incumbent in imperative languages, and encourages categorically-sound algorithms, while remaining performant thanks to LuaJIT's tail-call optimization.

As such, it is a better fit for core script environment code whose soundness (or lack thereof) waterfalls down to any scripts that run within it.

<a id="orgd75ac5d"></a>

## Won't multiple languages fragment the scripting ecosystem?

In theory, yes. In practice, the effect is likely amortized by several properties of the implementation and modding scene:

-   There are no barriers between languages in Scripting 2.0. Since all of them ultimately become Lua code, any functionality defined in one language can be accessed from another using standard Lua primitives.

-   As such, Fennel, Lua, and all other sub-languages are not black boxes to one another unless it is by design, so can (and in practice, do) import one another and share functionality.

-   Beyond the modded exes codebase itself, mods are separable to the point where users who don't want to see a `.tl`, `.fnl` or `.lua` file will never have to.

-   Freedom is a key part of Stalker modding; we're already up to our eyes in fragmentation of all kinds with no end in sight, to the point where soup mods embrace the idea and spur it full-speed into gleeful chaos

-   Positives include greater accessibility, better code (as a function of both language design and modder enthusiasm,) and accessibility-through-exposure

-   Negatives include making it harder to find support if you use a more niche language, and the ensuing faction wars between programming disciplines (which some may see as a positive, YMMV)

<a id="orga9e84db"></a>

## Not specifying extension in `require` or `import` is weird

Extension elision is standard Lua, as demonstrated by the default value of `package.path`, which searches for `?.lua` and `?/init.lua` by default.

While possible to implement by omitting extensions from `package.path`, the result would be less idiomatic, and reduce structural soundness by implicitly endorsing the coexistence of `<name>.script` with `<name>.lua`.

<a id="orgaeccce9"></a>

## Being able to import JSON and other data files as first-class scripts is weird

Perhaps, but convenient and categorically sound; as structured plain data, all such formats have a trivial mapping to Lua's table model.

Making them first-class obviates the reader <-> readee relation with regular scripts, as shared variable names provide a common through-line.

In addition, regular read-by-hand machinery remains available via Lua libraries, and is now easier to integrate by virtua of first-class `.lua`.

<a id="org1b8bb80"></a>

# Motivations

<a id="org78a36cf"></a>

## Selfish

I want to mod stalker in a well-featured language - specifically, Lisp - and organize my mods to the standards I've grown accustomed to in well-featured languages. I know Lua to be capable of this, so want to see it happen.

<a id="org110b126"></a>

## Altruistic

-   Moving engine surface area into a domain understood by more people facilitates new classes of mod, and makes it easier to contribute to a core engine system
-   Access to a first-class Lua environment opens up new organizational and computational possibilities for advanced Anomaly modders
-   Support for new scripting languages (some of which have overlap with other platforms, like LOVE and TIC-80) opens up Anomaly to more prospective modders, and facilitates both the accumulation and discharge of transferable skills
-   Use of `.lua`, `.fnl`, `.tl`, etc instead of `.script` allows better integration with editor syntax highlighting and the Language Server Protocol, resulting in a lower-friction development experience for modders willing to perform the associated setup.
-   Thinking longer-term, organization, redirection, and patching machinery makes it more feasible to organize / upgrade / refactor vanilla scripts

<a id="org068aef4"></a>

# Closing Notes

I've opened this PR as a draft, as it needs some final changes in order to fully fulfil the stated vision, but has now taken shape enough to be both recognizable by other modders, and runnable under standard Modded Exes packaging procedures.

(My rough TODO at time of writing can be found in [this gist](https://gist.github.com/ProfLander/1568088a0aee165c4db61be1f922d9bc).)

I thought it best to do so at this point, on account of its considerable scope; discussion, amends, and testing will no doubt be necessary before it can be merged in good conscience, so starting the process early is a beneficial move.

As the present generalized overhaul evolved from a simple special-case integration of `fennel.lua` via `lua-extensions` and baked executable resources, the commit history contains significant rewrites and some reverts. I've opted to leave this intact to transparently document its development history, rather than sanitize it into a series of redundancy-free feature commits.

Thank you for reading :)


nltp-ashes:
You should only add in the modded exes the baseline you need, and implement the rest in your own addons.

I am against the fact that you ported the modded exe's scripts to your new system.

This forces mod developers (who need to read modded exes files to understand how X thing was changed, or where new callbacks are being fired from, etc) and contributors to this repository to use your new system, whether they like it or not.

ProfLander:
Reproducing Discord discussion for clarity:

If you look at the changes, you'll find that much of the existing exes script code has been left alone in terms of form, if not in the way it's laid out on the disk. some was selectively reimplemented where it made sense (i.e. the options tree stuff that i'd already turned into a DSL has been further de-boilerplated,) but in broad existing code has simply split into multiple scripts and moved into nested directories.

The things that rely on my system are - in the main - new code comprising a Lua port of `CScriptEngine`, and benefit considerably from being implemented so, as detailed in the thesis above. Prime exceptions are `dxml` and `axr_main`, the entrypoint functions of which have been greatly simplified to use the new nesting-friendly recursive import machinery (which would also have been necessary for a simpler treatment as well, given the nature of folders-in-folders.)

Some patch scripts were renamed to Lua with minimal `local foo = require("foo")` and `return {}` wrapping for consistency with the surrounding code, but these cases could be reverted back depending on reception. In theory further could be Lua/Wua -fied with sufficient effort, but the benefits of doing so are questionable versus those that would be lost in the process.

themrdemonized:
I gotta be honest, I'm not ready to implement the changes of this size in the current version. Currently they go waaaay out of the scope of this project

ProfLander:
That's reasonable; this is a broad and far-reaching PR, not to mention balling up some fairly controversial changes / additions on top of more regular fare.

As such, I've been pulling it apart into chunks that might be more easily considered, the baseline of which is a move of `CScriptEngine` and related classes into Lua.

I'll put that up later, once I've had chance to write a digestible summary of it - could be that it's still too big in scope, in which case fair enough, but I may as well show what I've prepped.

For now, I'll close this one out.

PR #242, opened 2025-06-03
Fix black default crosshair

ProfLander:
Ensures crosshair color settings are initialized correctly on first boot, and reorders applicable color controls from ARGB to the more intuitive RGBA.


ProfLander:
Looks like I submitted as draft by accident - should be good to go.

PR #257, opened 2025-06-17
Spawn antifreeze

themrdemonized:
V3
* restored spawn_antifreeze functionality with edits, default is disabled
* Prefetch models in a separate thread before spawning in the main thread
*


themrdemonized:
Test release https://github.com/themrdemonized/xray-monolith/releases/tag/2025.6.17

Toggle with `spawn_antifreeze 1` in console or in Modded Exes options

fridvp:
had an issue with it
playing with mags redux and quite a lot of other mods. it seems to not spawn all the items it should (as in the items that spawn under same conditions with spawn_antifreeze 0) (talking about adding items to a trader). Also just gives a lot of microfreezes while spawning those items

ProfLander:
I played around with this in an attempt to get some thread profiling data, points of note:

- The new loading logic didn't seem to kick in when opening Sid's trade interface on a fresh Loner game, possibly due to items populating before / independently of the antifreeze system, but doing the same with Petrenko on a Duty start reliably triggers it.

- Visual Studio 2022's Concurrency Profiler extension worked once, producing a browseable thread graph, and then disappeared from the extensions menu forever; enable / disable / uninstall / reinstall do nothing. With any luck that's specific to my machine, since it seems to be path-of-least-resistance for threaded tracing.

- As an alternative, I tried AMD uProf, but it doesn't display the advertised per-thread flamegraph when attached to Anomaly; no related options in the UI, built-in presets make no difference, and the manual doesn't specify the preconditions for making it gather / process the necessary data to show it.

We may well end up needing IXRay-style Optick or some other integrated solution for consistent threaded profiling...

themrdemonized:
https://github.com/bombomby/optick bomboclat
https://github.com/ixray-team/ixray-1.6-stcop/commit/16e5dacc81515700094004e4665f05150212953d + example integration

themrdemonized:
V2:
https://github.com/themrdemonized/xray-monolith/releases/tag/2025.6.19

* Instead of device.seqParallel, use thread_spawn
* Prefetch models in separate thread before spawning game object, then spawn them normally
* Removed spawn_antifreeze batch variable, all events come and go same way as before

The idea is to reduce stutters by prefetching models and textures in the separate thread before spawning. I removed batch processing, so scripts and other parts of engine should work correctly as before. Trader items might get freezes back, but i think its bearable

ProfLander:
Bomboclaat actioned: #262 

Should help with figuring this out!

themrdemonized:
изображение
So there is a noticeable freeze that happens 2 frames before game object spawn kicks in and subsequent prefetching. The freeze happens when trying to spawn an NPC in particular While prefetching does save some time, its not the main cause for a freeze. Need investigating

themrdemonized:
V3: https://github.com/themrdemonized/xray-monolith/releases/tag/2025.6.21

This one looks pretty good, works with actual visuals of alife objects. Need to test for consistency and stability.

As the sidenote of this feature, any scripts that spawn alife objects and expect game_object on next frame won't work correctly. It usually was the case for vanilla code as well but its guaranteed with this feature.

themrdemonized:
изображение
`army_sim_squad_10` squad
900 ms loading stutter is offloaded, actual spawning is 93 ms

LVutner:
peak modding

PR #264, opened 2025-06-21
CMake Integration

ProfLander:
Integrates the CMake build system with Anomaly, in-line with OpenXRay and IX-Ray.

## Reasoning

- Free choice of code editor
  - CMake supports generating project files for VS 2015, 2017, 2019, 2022
  - Later VS versions directly support CMake projects
  - CMake is supported via plugin for editors like VS Code, Emacs, Vim, etc.

- Free choice of compiler
  - MSBuild / GCC / Clang
  - Non-MS compilers are known to produce better-optimized machine code
  - Access to better parallel build tooling like Ninja for faster builds

- Easier to maintain
  - `CMakeLists.txt` is more git-friendly and explicit than VS' GUI-driven project configuration + corresponding markup data
  - Scriptable builds mean better automation and features
    - ex. `ccache` for build artifact caching
  - More flexible structure should help solve the existing combinatorial explosion of build configs

## Tasks

- [x] Compile all modules in isolation
- [x] Compile and link runnable executable via `xrEngine`
- [x] Integrate `rc` files for splash screen and program icon
- [x] Integrate `mt` for applying program manifest
- [x] Reimplement solution folder structure
- [x] Add all gamedata via dummy cmake targets
- [x] Integrate with VS' native CMake handling
- [x] Bump C++ edition to 14
- [x] Expose all internal libraries as modules
- [x] Reintegrate existing DX / AVX build configurations as executable targets
  - [x] Enable / disable exceptions based on build type
  - [x] Strip redundant preprocessor definitions
  - [x] Reimplement AVX compile flag
  - [x] Reimplement other compile flags
- [x] Linkage pass for libraries
  - Need to explicitly specify dependencies via `target_link_libraries`
  - Use `PUBLIC` to propagate include paths
  - Use `PUBLIC` to propagate compiler flags (ex. `USE_VERIFY_IN_RELEASE`)
- [x] Formalize `StackWalker` module
  - Currently globbed into another module, failing to generate PDB
- [x] Fix DX10 build under new linkage 
- [x] Polish remaining inter-module links
- [x] Module naming pass
- [x] Strip out redundant VS project files
- [x] Fix `xrGame` exceeding `/GL` image size limit
- [x] Split `xrUI` and `xrScripts` out of `xrGame` as per IX-Ray
- [x] Reintegrate Profiled as a build configuration
- [x] Reintegrate Verified as a build configuration
- [x] Fix failure to link `luabind` in `Verified` configuration
  - Not `luabind`; was caused by C++ dynamic initializers for `xrCore` and `xrRender_R*` running out-of-order, resulting in a call to a null `Memory.mem_fill`. Fixed by explicitly linking `xrCore` in `xrEngine`-derived targets.
- [x] Consider replacing `msvc.cmake` with [WindowsToolchain](https://github.com/MarkSchofield/WindowsToolchain)
- [x] Integrate `ninja`
- [x] Variable usage pass
  - [x] Implement cached configuration
  - [x] Reduce code duplication
- [x] Custom target for building gamedata
- [x] Strip out redundant batch files
- [x] Install step for staging exes, gamedata, PDBs
- [x] Namespacing for `cmake/*` modules
- [x] Profile Ninja against MSBuild
- [x] Integrate with GitHub Actions
- [x] Update `README.md`
- [x] Second-pass `vcxproj` -> CMake config translation
  - [x] Add header files to sources
  - [x] Recapture VS exclusions
- [x] Fix failure to find `mt` in buildbot environment
  - Need to investigate MSBuild setup action
    - See if there's some param or alternative to allow `mt` access
- [x] Undraft

## Further Work

- Troubleshoot `dog_state_manager.cpp` rebuild
  - Unnecessarily marked as changed when rebuilding with no edits
  - Other files exhibited this, but had obvious malformed `#include`s with spaces before or after the path; not so for `dog_state_manager.cpp`
  - Invokes extra link time during CI

- Integrate `clang`

- Integrate `gcc`

- Reorganize directory structure
  - Old VS source groupings should be recaptured as actual directory structure, and split out into static libraries where appropriate
    - CMake integrations outside of static VS solution generation don't have the concept, rely on portable directory approach

- Setup Linux building
  - Full native Linux requires SDL / OpenGL integration, intractable in the short term
  - Should be able to build with MSYS2 and link against wine libraries to produce a wine-runnable `.exe` and enable development on Linux

- Investigate splitting various subsets of `xrCore`, `xrEngine`, `xrGame` into their own modules
  - `xrEngine` and `xrGame` are particularly large
    - Results in longer compiles, as entire modules recompile on change
  - Are there static analysis tools that can determine inter-file dependencies?
  - Likely various redundant code also, ex. GameSpy

- Investigate merging `xrRender_` into a single project
  - Eliminate duplicate code
  - Gate remaining differences behind preprocessor blocks
  - Alternately: Investigate dynamic libraries so `Anomaly.exe` and `AnomalyAVX.exe` can be shipped alongside a set of `xrRender_R*.dll`s, load them dynamically at boot based on a CLI argument

- Fix debug build
  - Currently has various compile errors due to outdated code
  - Some of these have been fixed, but the remainder are beyond scope of this PR


damoldavskiy:
Really nice! Looking forward to be able to uninstall VS

LVutner:
peak modding

ProfLander:
Good to go.

@themrdemonized Do you think it's time to consider a `master` branch or similar, now the codebase is vs2022-agnostic?

themrdemonized:
The master branch is linked to bitbucket origin of vanilla anomaly, i want it to keep that way for now

ProfLander:
Ah, fair enough - I'll leave you to figure out appropriate nomenclature :)

themrdemonized:
~~Please update README.md for how to use CMake, even i still struggle with it. Like briefly how to generate solution for VS2022 as a default choice~~

I see already done, my bad

PR #271, opened 2025-06-28
Don't apply rendering optimizations to cars and helicopters

OneMorePseudoCoder:
Исправляет невидимые машины (и по большей части вертолёты) при включённой оптимизации геометрии.

### DIFF
--- a/src/xrGame/Car.cpp
+++ b/src/xrGame/Car.cpp
@@ -205,6 +205,18 @@ BOOL CCar::net_Spawn(CSE_Abstract* DC)
 		m_memory = xr_new<car_memory>(this);
 		m_memory->reload(pUserData->r_string("visual_memory_definition", "section"));
 	}
+	
+	renderable.visual->flags.set(IRenderVisualFlags::eIgnoreOptimization, TRUE);
+
+	xr_vector<IRenderVisual*>* children = renderable.visual->get_children();
+
+	if (children)
+	{
+		for (auto* child : *children)
+		{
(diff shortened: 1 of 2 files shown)


solbjorn:
@OneMorePseudoCoder, was this purely your idea or you saw that in my fork / new upstream OGSR? It's somewhat strange you reached out to me with some interest to my code and then this PR appears.

I mean, you can take whatever you want, but I'd love to see the original authorship / credits then.
I mean, just in case. This could be just a coincidence.

OneMorePseudoCoder:
@solbjorn Oh, sorry man. Yeah, I took inspiration from your fork. I played Anomaly recently with geometry optimization turned on and noticed that all helis look messed up. I instantly remembered that you had fix to this and also that birds have optimization forcefully turned off in code. So I simply opened ai_crow.cpp file, copied some code, and pasted into Car.cpp and Helicopter.cpp.
Next time I will give you a credit even if I just take inspiration from you code.

solbjorn:
@OneMorePseudoCoder, ok. BTW the idea comes from the OGSR team, not me.

Also I think that in case of hierarchy visuals you don't need to traverse all the children and set optimization ignore flag for each of them (it's enough that the root visual has this flag). Have you tried just setting this flag to the base visual without iterating over the children?

themrdemonized:
Can you investigate strange bug ? It happens with conjuction with this mod https://www.moddb.com/mods/stalker-anomaly/addons/new-walk-run-animation
изображение

PR #277, opened 2025-07-01
CMake Refinements

ProfLander:
- Fix statically-linked ICU libs adding ~26.7Mb to final binary size
- Add missing compiler flags
- Deduplicate existing linker flags
- Remove redundant PCH machinery
- Fix incorrect `xrCDB` PCH definition
- Misc CMake robustness tweaks and tidying


ProfLander:
Sorry, forgot to switch branch before continuing on to module structuring - have rewound this one back to the original set of commits to make it mergeable, and will continue with further changes on a temporary one.

PR #279, opened 2025-07-02
CMake: Modules, Clang, Buildbot

ProfLander:
Since CMake's source grouping feature only works with statically-generated VS projects, the original `.sln` organization needs to be reinstated using cross-platform machinery.

Reasoning:
- CMake has a concept of folders over targets (libraries, executables) but leaves folders within them (containing cpp / h files) to the filesystem
- Ideally we don't want to change the existing FS layout too much right now, if at all, for the sake of remaining comparable to OXR and other forks, and to avoid disrupting open PRs
- Ergo, the existing leaf-level folder structure needs to be split into static libraries and organized using CMake folders
- Additional wins on the compiler side, as smaller modules means recompiling less code on change once linkage is separated from their existing top-level module

Additional:
- Support for the `clang` compiler
  - A step toward cross-platform support
  - Produces binaries, but they currently crash due to X-Ray's reliance on MSVC extensions

- Buildbot refactor
  - Separate workflows:
    - Development
      - Builds one configuration on each supported compiler
        - Ensures new code complies with standard C++
        - Defaults to AVX configure presets, `Verified` build preset, `Anomaly.DX11` target
        - Can be controlled via fork-local GitHub environment variables
      - Designed for quick developer feedback when developing local forks
    - Pull Request
      - Runs on any pull request
      - Defaults to `RelWithDebInfo` build preset
      - Runs the full compiler / renderer / AVX build matrix
      - Designed to thoroughly verify code before merge
    - Release
      - Runs when committing to the `cmake` branch
      - Same as Pull Request, but publishes a release on successful build
      - Separation from PR workflow avoids marking mergeable code as failed due to attempting an illegal release step

- [x] Simplify module definition via macros
- [x] Separate `XRay.Engine` structure from executables to allow restructuring
- [x] Address AVX
  - Needs to apply to all libraries, not just the final .exe
    - Ergo, probably needs to be a toolchain-level setting
    - i.e. `MSVC.MSBuild.AVX` etc.
- [x] Address long generation times, large compile intermediates with many modules
  - Need to make modules interfaces by default, gather together into actual top-level libraries
  - [x] Deprecate `CHILD_OF` semantic in favor of path analysis
  - [x] Make `INTERFACE` default, override at top-level modules

- [x] Recapture original solution grouping via CMake targets
  - [x] xrAPI
  - [x] xrCDB
  - [x] xrCore
  - [x] xrCPU_Pipe
  - [x] xrEngine
    - [x] Engine
    - [x] Game API
    - [x] General
    - [x] Interfaces
    - [x] OpenAutomate
    - [x] Render
    - [x] RenderRef
  - [x] xrGame
    - [x] AI
    - [x] Core
    - [x] Physics
    - [x] UI
  - [x] xrNetServer
  - [x] xrParticles
    - [x] Kernel
    - [x] PAPI
  - [x] xrPhysics
    - [x] Interface
    - [x] Kernel
    - [x] Physics
  - [x] xrRender_R1
  - [x] xrRender_R2
  - [x] xrRender_R3
  - [x] xrRender_R4
  - [x] xrSound
    - [x] Cache
    - [x] Core
    - [x] Emitter
    - [x] Environment
    - [x] Kernel
    - [x] Source
    - [x] Target
  - [x] xrXMLParser
    - [x] Kernel
    - [x] TinyXML
 
- [x] Look into handling git submodules via CMake
  - Prevents user error
  - Allows submodules to be cloned selectively by need
- [x] Faster builds for `Profiled`
  - [x] Remove `/LTCG` and `/GL`
  - [x] Try incremental linking

Future work:

- PCH pass
  - Need to convert inline `stdafx.h` etc over to CMake's shareable system, simplify existing dependencies

- Linkage pass
  - Need to link submodules directly where possible, instead of gating them behind their parent module


ProfLander:
Alright, this has been merged up to the latest `all-on-one-vs2022-wpo`, so should be good to go.

It's set up assuming `cmake` is set as the main branch; that's controlled by the top of `anomaly-development.yml` and `anomaly-release.yml` in case you want to do any renaming. Beyond that, kick the buildbot with a commit to the main branch after switching over and it should deploy the inaugural CMake release.

PR #288, opened 2025-07-09
Added a link to script debugger section

SaloEater:
Need this link to refer people directly to the instructions about script debugger i.e. http://github.com/saloeater/xray-monolith#script_debugger_instructions


themrdemonized:
I think a good idea would be to contribute a page to anomaly modding book

SaloEater:
Right, completely forgot about that, but modded exes is still the only source of truth we all can rely on so it would be good to have small description here as well

PR #306, opened 2025-07-26
Model Visbox Update and Script Attachment Fixes

xr-lucy:
- Fix script attachments breaking existing model callbacks (used by attachable items like headlamps)
- Moved update of hud item visbox to CHudItem and disabled it by default (uncomment in HudItem.h if there's a need for it)
- Force update bones and visbox of script attachment models when bone visibility is changed


LVutner:
image

PR #310, opened 2025-07-30
CCar CWeaponStatMgun changes

GhenTuong:
- CCar: Support flying physic to make drones.
- CWeaponStatMgun: Further development for this unused class. Make it useable, make server object, save ammo count, multi-barrels gun, stalkers using object.
- CHolderCustom: Improve holder interaction for actor. Attach, detach, UI ammo count.


Bence7661:
For these type of PRs a link to a YT video showcasing stuff would be cool.

PR #318, opened 2025-08-07
Dynamic Devices turn on/off

Ncenka:
Allow to dynamically turn on/off devices by `set_device_enabled(bool)` and check if device is turned on/off by `is_device_enabled()`.


themrdemonized:
Please add new functions into lua_help_ex

Ncenka:
> Please add new functions into lua_help_ex

Done

PR #323, opened 2025-08-08
LuaJIT vs2022 project

themrdemonized:
Moves LuaJIT from NMake procedure to VS2022 project
batch_build.bat is now broken for some reason
msbuild console app is also broken


themrdemonized:
fixed, testing for release

PR #326, opened 2025-08-09
Reapply "Launchers: Cartridge Ammo + Trajectory and Reload Options" 

ProfLander:
Fixes the M79 / RG6 bugs introduced in the original PR.

#### Summary of bug
- Inadvertently moved `CWeaponMagazinedWGrenade`'s ammo decrement into `CWeaponGrenadeLauncher`, causing it to apply to `CWeaponRG6`, which already defers to `CWeaponShotgun` for that functionality
- As a result, M79 was trying to decrement its ammo pool twice, causing the initial launch to work, but immediately canceling all followup actions (anim, sound)
- RG6 was doing the same, but succeeding when the ammo count was >=2; same behaviour as M79 when firing with 1 shell loaded

### DIFF
--- a/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -600,6 +600,23 @@
 		<text>Gameplay</text>
 	</string>
 
+	<!-- Gameplay General -->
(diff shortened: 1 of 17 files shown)


ProfLander:
Removed some CMake leftovers from the commit log; should be good now.

PR #337, opened 2025-08-15
Add string count to `stat_memory` and OOM handler

ProfLander:
- Thread a `u32` reference through `stat_economy` to avoid adding a global
- Rearrange existing `Msg` calls to segment `string` and `shared`, add count to `string`

### DIFF
--- a/src/xrGame/console_commands.cpp
+++ b/src/xrGame/console_commands.cpp
@@ -336,7 +336,7 @@ static void full_memory_stats()
 	Msg("* [x-ray]: process heap[%u K], game lua[%d K], render[%d K]", _process_heap / 1024, _game_lua / 1024, _render / 1024);
 #endif // SEVERAL_ALLOCATORS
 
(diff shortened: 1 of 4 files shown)


src/xrGame/console_commands.cpp
```diff
@@ -335,7 +336,8 @@ static void full_memory_stats()
 	Msg("* [x-ray]: process heap[%u K], game lua[%d K], render[%d K]", _process_heap / 1024, _game_lua / 1024, _render / 1024);
 #endif // SEVERAL_ALLOCATORS
 
-	Msg("* [x-ray]: economy: strings[%d K], smem[%d K]", _eco_strings / 1024, _eco_smem);
+	Msg("* [x-ray]: strings: memory[%ld K], count[%lu]", _eco_strings / 1024, _eco_strings_count);
```
Xottab-DUTY: Maybe this?
```suggestion
	Msg("* [x-ray]: shared strings: memory[%ld K], count[%lu]", _eco_strings / 1024, _eco_strings_count);
```

ProfLander: Sure - more explicitness sounds good.

PR #338, opened 2025-08-16
Added the new "volume_mult" property for HUD sound call back. Removed all indoor framework related engine side code

Kutez:
Add the new `volume_mult` for `on_before_play_hud_sound` callback. You can use it to change the volume of any hud sound, with or without replacing the sound itself. (more info in the `callbacks_gameobject.script`)
Example:
```
	if alias == "sndShot" and parent and parent:id() == AC_ID then
		result.section = "wpn_l96a1_sounds"
		result.line = "snd_shoot"
		result.volume_mult = 0.75 -- Multiply the new fire sound's volume by 0.75
	end
```
Or:
```
	if alias == "sndShot" and parent and parent:id() == AC_ID then
		result.volume_mult = 0.5 -- Multiply the old fire sound's volume by 0.5
	end
```

Also removed all indoor framework related engine side code become the project the move to Lua and rely on the callback.


Kutez:
Sorry for the long ass title lol

Kutez:
Known bug, `::luabind::object_cast<float>` return nil when the value is 0 for some reason, others worked fine.

themrdemonized:
By the looks of it should work fine. Maybe you can set the default value of volume multiplier to 1 instead of nil, that way you wouldn't need a nil check

Kutez:
It still doesn't solve the problem of when the property is 0, `::luabind::object_cast<float>` will just return nil. I will do some hack to see if it works

Kutez:
Ok I was wrong, it does return 0 but idk that in C++, not (a 0 float value) return true

Kutez:
Apparently, a 0.f is the same as NULL, so there is no way afaik I can distinct them, I will just put a notice. God this driving me crazy.

src/xrGame/HudSound.cpp
```diff
@@ -309,16 +309,24 @@ void HUD_SOUND_COLLECTION_LAYERED::PlaySound(LPCSTR alias, const Fvector& positi
 			::luabind::object output = funct(alias, parent_lua_game_object);
 			if (output && output.type() == LUA_TTABLE)
 			{
+				float volume_mult_ex = ::luabind::object_cast<float>(output["volume_mult"]);
 				LPCSTR section = ::luabind::object_cast<LPCSTR>(output["section"]);
 				LPCSTR line = ::luabind::object_cast<LPCSTR>(output["line"]);
+				if (volume_mult_ex) {
```
Xottab-DUTY, line 318:
```suggestion
				if (!fis_zero(volume_mult_ex))
				{
```

src/xrGame/HudSound.cpp
```diff
@@ -309,16 +309,24 @@ void HUD_SOUND_COLLECTION_LAYERED::PlaySound(LPCSTR alias, const Fvector& positi
...
+					volume_mult = volume_mult*volume_mult_ex;
+				}
 				if (!section)
 				{
-					Msg("!_G.COnBeforePlayHudSound callback, HUD_SOUND_COLLECTION_LAYERED::PlaySound, failed to override sound item %s, no section specified", alias);
-				} 
+					if (!volume_mult_ex) {
```
Xottab-DUTY:
```suggestion
					if (fis_zero(volume_mult_ex))
					{
```

src/xrGame/HudSound.cpp
```diff
@@ -309,16 +309,24 @@ void HUD_SOUND_COLLECTION_LAYERED::PlaySound(LPCSTR alias, const Fvector& positi
...
+						Msg("!_G.COnBeforePlayHudSound callback, HUD_SOUND_COLLECTION_LAYERED::PlaySound, failed to override sound item %s, no section specified", alias);
+					}
+				}
 				else if (!line)
 				{
-					Msg("!_G.COnBeforePlayHudSound callback, HUD_SOUND_COLLECTION_LAYERED::PlaySound, failed to override sound item %s, no line specified", alias);
-				} 
+					if (!volume_mult_ex) {
```
Xottab-DUTY:
```suggestion
					if (fis_zero(volume_mult_ex))
					{
```

src/xrGame/HudSound.cpp
```diff
@@ -353,7 +361,7 @@ void HUD_SOUND_COLLECTION_LAYERED::PlaySound(LPCSTR alias, const Fvector& positi
 
 	for (; it != it_e; ++it)
 	{
-		if (it->m_alias == alias_to_play)
+		if (it->m_alias == alias_to_play && volume_mult>0)
```
Xottab-DUTY:
```suggestion
		if (it->m_alias == alias_to_play && volume_mult > EPS_S)
```

src/xrGame/WeaponMagazined.cpp
```diff
@@ -112,8 +109,8 @@ void CWeaponMagazined::Load(LPCSTR section)
 	m_sounds.LoadSound(section, "snd_shoot", "sndShot", false, m_eSoundShot);
 	if (WeaponSoundExist(section, "snd_shoot_actor"))
 		m_sounds.LoadSound(section, "snd_shoot_actor", "sndShotActor", false, m_eSoundShot);
-
 	//-Alundaio
+
```
Xottab-DUTY:
```suggestion
```

Xottab-DUTY (commented):
Comparing floats with `0` directly is potentially erroneous/inaccurate due to how floats work in hardware.
X-Ray has a special function `fis_zero` to check if a float is zero and it's always better to use it.

> 0.f is the same as NULL

It is not, but luabind automatically converts `nil` to `0.f`. It's possible to avoid that.

Xottab-DUTY:
> It's possible to avoid that.

To detect if Lua's `volume_mult` is actually `nil`, don't do `luabind::object_cast` immediately.
Here's an example:
```cpp
auto volume_mult_ex_obj = output["volume_mult"]; // receive Lua object
bool volume_mult_ex_is_nil = luabind::type(volume_mult_ex_obj) == LUA_TNIL; // the actual check for nil
float volume_mult_ex = ::luabind::object_cast<float>(volume_mult_ex_obj); // receive the actual value. nil will be converted to 0.f automatically
```
Then, you can use `volume_mult_ex_is_nil` in places where `fis_zero` is not enough and you need exactly nil checking.

themrdemonized:
let me know when its ready for review

Kutez:
@themrdemonized ready for review, thanks

themrdemonized:
I made a commit that simplifies the `volume_mult` logic https://github.com/themrdemonized/xray-monolith/commit/b39e5c55b905d4c6e0ccc81eca39cc367b50d50a

PR #350, opened 2025-08-26
On use functor callback

antglobes:
Get the gameobject, functor slot and name after clicking an action in the item properties box, inside of the player's inventory.


antglobes:
Made mistake, didn't test hang on

PR #353, opened 2025-08-27
Level Script Attachments

xr-lucy:
Added support for level script attachments (same function calls as with objects, just level.x instead of obj:x)
For example:
```lua
local latt = level.add_attachment("gubu", "dynamics\\scene_objects\\pripyat\\pri_a25_antenna_destroyable_barrier")
```

To attach an existing attachment to the level use
```lua
attachment:set_parent_level()
```


themrdemonized:
lmk when its ready

PR #354, opened 2025-08-28
ImGui: Implement grouping API

ProfLander:
Adds an API for registering widgets to globally-named groups, and rendering them in order of registration.

Allows mods to extend the built-in ImGui overlay in a structured fashion, and expose extension points where other mods may inject their own UI elements. The main API is given in `imgui_groups.script`, and contains full documentation, usage instructions, and a basic example.

Adds the following groups by default:
- `Unique` - Renders unconditionally
- `Main` - Renders when the ImGui overlay is visible
- `MenuBar` - Renders elements within the built-in overlay menu bar
- `Debug` - Built-in submenu, intended for debugging tools like inspectors, editors, stats, etc.
- `Mods` - Built-in submenu, intended for simple mod interfaces that don't necessitate a dedicated dropdown


ProfLander:
CC @xr-lucy for review 🙏

xr-lucy:
I think it should be kept in a single file (i know there are multiple files as a way to demonstrate how to use it but it feels like unnecessary bloat to add 4 files for the API)

Otherwise it's a great idea 😄 

I guess you could even add the functions to _G or ImGui namespace, maybe something like ImGui.AddGroup("name") or (in _G) AddImGuiGroup("name") similar to the way script callbacks are added and registered, but it's up to you 🐱

ProfLander:
Fair point on files - especially since the group addition code is minimal; I've moved it into a dedicated subheading in `imgui_groups.script`.

Namespace-wise, I think `_G` is already too busy, and `ImGui` hosting C++ methods is a nice parallel with the official manual, so I went with `ImGui.Groups` to keep the footprint minimal, and upgraded the function factory to create sub-namespaces:

```lua
ImGui.Groups.Add("MyGroup") -- Creates ImGui.Groups.MyGroup
ImGui.Groups.MyGroup.Widget(my_widget) -- Add a widget to it
ImGui.Groups.MyGroup.Render() -- Render it
```

PR #364, opened 2025-09-09
Fix `CALifeMonsterBrain::process_task` segfault

### LINKED ISSUE #363: Hidden Error: Segmentation fault in `CALifeMonsterBrain::process_task`
`CALifeMonsterBrain::process_task` periodically causes a segmentation fault during offline A-Life update.

This is not noticeable under normal circumstances, as the segfault is rendered catchable by MSVC's structured exception handling, and caught by [luabind's built-in `try` / `catch`](https://oberon00.github.io/luabind/errors.html#structured-exceptions-msvc) which bridges C++ callbacks to the C Lua runtime.

The resulting stack unwind causes the remainder of the monster brain's update to be skipped, likely hampering offline interactions for affected objects. In my testing, this has occurred with generic simulation stalkers and zombified, with space restrictors around the clear sky base and miracle machine entrance yielding the problematic null tasks on a new game - no instances of successful runs have been recorded so far.

Compiling with Clang exposes the error, as it [does not deal in structured exceptions](https://github.com/llvm/llvm-project/issues/62606), but appears otherwise playable following recent fixes to [ProfLander/xray-monolith:cmake](https://github.com/ProfLander/xray-monolith/tree/cmake).

(I tried the `-fasync-exceptions` flag mentioned in the linked issue: it causes an ICE when attempting to build `xray-monolith`, and is ultimately a workaround for the root issue of null tasks.)

The relevant code is located [here](https://github.com/themrdemonized/xray-monolith/blob/20b9d6a482cd309c7d64ce23686ab790c0d0e225/src/xrServerEntities/alife_monster_brain.cpp#L115): `task` is assigned `nullptr`, and the resulting dereference throws the segmentation fault.

## Known Details

Investigating the LTX sections for the affected restrictors doesn't turn up anything out-of-place.

When accessed from `CALifeMonsterBrain`, the `task()` invocation goes to the virtual base method defined on `CSE_ALifeSmartZone`, which returns `0`. This may be accounted for by the absence of a `task` implementation on `cse_alife_smart_zone` (declared in `lua_help.script`, bound to `CSE_ALifeSmartZone` engine-side) or its inheritor `se_smart_terrain` (defined in `smart_terrain.script`.)

However, the smart terrains / space restrictors in question appear to yield a valid task when accessed from elsewhere in Lua via `se_smart_terrain:get_alife_task()` - I successfully monkey-patched this method with logging, but ~~attempting the same with `se_smart_terrain:task()` resulted in no change.~~

**Edit:** User error, patch worked. Fix PR sent.

I've looked over OpenXRay and IX-Ray, and the relevant code (absence of relevant gamedata notwithstanding) appears functionally equivalent. Given that they maintain Clang and GCC builds which would see the same result at runtime, and don't appear to be using any special compiler flags to sidestep the issue, it's reasonable to assume that they don't suffer from the underlying bug.

## Reproduction Save

[monster-brain-segfault-save.zip](https://github.com/user-attachments/files/22226942/monster-brain-segfault-save.zip)

Should be vanilla-compatible; made on a local development preset with no save / simulation -affecting mods loaded. Reproduces on `all-in-one-vs2022-wpo` mainline as well as on MSVC / Clang presets via my `cmake` branch.

Spawn at the southeast entrance to Garbage, and walk to the pair of stalkers stood just beyond the nearby campfire. Once they finish greeting you, the game should have had time to trigger the error. Otherwise, reload and try again - it can occur before or slightly after.

In theory this should show in the log on a Verified build as a result of the `THROW3` expression, but that doesn't seem to happen; adding a conditional `if (task) { ... } else { ... }` with `Msg` logging inside `CALifeMonsterBrain` and setting breakpoints is sufficient to demonstrate the problem on MSVC.

ProfLander:
Adds a working implementation of `se_smart_terrain:task()`, which fixes `CALifeMonsterBrain::process_task` segfaulting after attempting to dereference the null pointer returned when falling back to the default virtual implementation of `CSE_ALifeSmartZone::task`.

As a result, offline monsters (+ stalkers) attached to a smart zone will now pathfind toward it instead of getting stuck in a broken state.

Closes #363.

### DIFF
--- a/gamedata/scripts/aaaa_script_fixes_mp.script
+++ b/gamedata/scripts/aaaa_script_fixes_mp.script
@@ -607,3 +607,11 @@ function trans_outfit.transparent_gg()
 		end 
 	end
 end
+
+-- Lander:
+-- Implement missing `CSE_ALifeSmartZone::task()` bind.
+-- Fixes segfault inside `CALifeMonsterBrain::process_task`,
+-- which was preventing offline monsters (+ stalkers) from pathfinding.
+function smart_terrain.se_smart_terrain:task()
+	return self.smart_alife_task
+end


themrdemonized:
Please put the code into `aaaa_script_fixes_mp.script` file

ProfLander:
And done 👍

PR #368, opened 2025-09-11
Fix options menu memory access violation on localization change

### LINKED ISSUE #367: Hidden Error: Memory access violation in options menu during localization reload
When the language is changed via the options menu, the localization reload function causes the main menu to be `xr_delete`'d and replaced with a fresh copy before option callbacks have finished running.

As the `UIOptions` Lua class accesses `self.owner` - a main menu reference populated at construction time - this causes MAVs when the resulting pointer is accessed in post-callback logic, which are hidden by luabind's SEH-empowered exception handler under MSVC.

There are two distinct variants of MAV, one involving access to `self.owner`'s child `CUIMMShniaga`, and the other to `self.owner` itself. Which occurs is inconsistent, though `Crash 1` can be caused consistently by inspecting deep into the call stack with the debugger. A non-debugged build appears to pick randomly between it and `Crash 2`, potentially implicating GC as cause of the variation.

Curated call stack:
```c++
<Main Menu>
<Options>
<Change Language>
<Apply>
// Button handler
CUI3tButton::OnClick()
    // Dispatch to lua
    CUIDialogWndEx::SendMessage(CUIWindow* pWnd, s16 msg, void* pData)
        // Options accept handler
        ui_options.UIOptions:On_Accept()
            // Localization update functor
            ui_options.func_localization()
                // Script loctable reload
                game.reload_language()
                    // Loctable reload
                    CStringTable::ReloadLanguage()
                        // Hide main menu
                        CMainMenu::Activate(false)
                        // Show main menu
                        CMainMenu::Activate(true)
                            // Reinstantiate start dialog
                            CMainMenu::ReloadUI()
                                // Construct new start dialog
                                NEW_INSTANCE(TEXT2CLSID("MAIN_MNU"))
                                    // Lua constructor, sets `owner` on options menu
                                    ui_main_menu.main_menu:__init()
                                        // Build controls
                                        ui_main_menu.main_menu:InitControls()
                                            // Initialize shniaga + magnifier
                                            CScriptXmlInit::InitMMShniaga("shniaga_wnd",self)
                                // Delete old main menu
                                xr_delete(m_startDialog)
            // Accept handler forwards to cancel handler
            UIOptions:On_Cancel()
                // Attempt to show the main menu this options menu was constructed with
                // (The `m_startDialog` that was `xr_delete`'d above)
                self.owner:ShowDialog(true)
                    // Dialog window show handler
                    CUIDialogWnd::ShowDialog
                        // Dialog holder start
                        CDialogHolder::StartDialog
                            <Crash 1 - access violation, self.owner has been deleted>
                // Attempt to hide this options menu (child of `m_startDialog`)
                self:HideDialog()
                // Attempt to call the main menu's lua-level show function
                self.owner:Show(true)
                    // Lua show function tries to show the magnifier
                    ui_main_menu.main_menu:Show()
                        // C++ handler
                        CUIMMShniaga::SetVisibleMagnifier(true)
                            <Crash 2 - access violation, self has been deleted>
```

ProfLander:
Adds a monkey-patch for `ui_options.script` to prevent it from accessing a deleted `self.owner` after localization is reset.

Closes #367.


gamedata/scripts/aaaa_script_fixes_mp.script
```diff
@@ -615,3 +615,52 @@ end
...
+local loc_reloaded = false
+
+--- Patch `func_localization` to set the reload flag
+local old_funct = ui_options.func_localization
+function ui_options.func_localization(...)
+    loc_reloaded = true
+    return old_funct(...)
+end
+
+--- Replace the copy emplaced in `ui_options.options`
```
ravenascendant: This should not be needed if the patch happens early enough.  the options table is built by ini_opt_base and may be empty when this code runs.  as long as the patch happens before init_opt_base runs this will be unnecessary.

gamedata/scripts/aaaa_script_fixes_mp.script
ravenascendant, line 1: the patch is running too late and shouldn't be in this file as this file is not loaded at the first main menu. 

this fix should be in _g_patches.script in order to be loaded before the first main menu. Alternatively a modxml script will also load before the ui_options.script is loaded by the main menu.

ravenascendant (changes requested):
patch timing is wrong. option table navigation is unnecessary if the functor is patched before the table is built.

themrdemonized:
> patch timing is wrong. option table navigation is unnecessary if the functor is patched before the table is built.

What you think if we put this change into ui_options.script itself and add this into the next exes version?

ravenascendant:
> > patch timing is wrong. option table navigation is unnecessary if the functor is patched before the table is built.
> 
> What you think if we put this change into ui_options.script itself and add this into the next exes version?

Are any of the pre MCM mods that replaced UI options still a thing? warfare overhaul? dynamic mutants? 

other than the risk of being over written, which is probably small (?) moving all your options changes to a script replacement does make sense.

gamedata/scripts/aaaa_script_fixes_mp.script
ProfLander: Good catch; I was testing on a straight-to-map preset and overlooked this - have moved the logic to `_g_patches.script`.

gamedata/scripts/aaaa_script_fixes_mp.script
ProfLander: Confirmed locally, it works without the `ui_options.options` manipulation following the move to `_g_patches`.

ProfLander:
Adopting `ui_options` into modded exes sounds good to me; I've drafted the PR while the question is open.

@ravenascendant How extensive are MCM's changes to `ui_options`? If `xray-monolith` is going to adopt the script as first-class gamedata, it would be preferable for that version to be the single source of truth, with downstream additions coming in via monkey patch.

We could break apart larger functions, or add new patch entrypoints within them if the current structure is too monolithic for that to be tractable as-is. Though I suppose that would raise the question of compatibility with vanilla 🤔

ravenascendant:
> Adopting `ui_options` into modded exes sounds good to me; I've drafted the PR while the question is open.
> 
> @ravenascendant How extensive are MCM's changes to `ui_options`? If `xray-monolith` is going to adopt the script as first-class gamedata, it would be preferable for that version to be the single source of truth, with downstream additions coming in via monkey patch.
> 
> We could break apart larger functions, or add new patch entrypoints within them if the current structure is too monolithic for that to be tractable as-is. Though I suppose that would raise the question of compatibility with vanilla 🤔

MCM doesn't use anything in ui_options.script 

ui_mcm.script started as a lazy copy and rename of ui_options.script with a dynamic options table loader tacked on. I have since expanded upon it in many ways.  "lazy" code duplication saves me from being affected by changes to ui_options.script
However bugs involving ui_options are likely to behave similarly with MCM because the structure and a majority of the code is still the same. 


it uses some of the same textures and texture definitions. and some translation strings. eventually i will duplicate these into MCM since i know that Tronex made a new options UI for anomaly 1.6 and i don't want to be dependant on file that could be deleted or changed becaue of that. 

if ui_options does become part of modded exe's i will submit a PR for the ui_options bugs that i have fixed in MCM. maybe open an issue for the ones i can't figure out how to fix. ;P

ProfLander:
> MCM doesn't use anything in ui_options.script

Gah, sorry; I got `ui_main_menu.script` mixed up with `ui_options.script`. Ran into an MO2 conflict with it in the early stages of debugging this PR, and the wrong one stuck in my head. No worries on that front then.

PR #394, opened 2025-12-01
Export Screenshot Func + variable resolution & encoding

antglobes:
- Extends level class to take screenshots in dds format
- Most common encoding formats included as global enum
- Take any filename written to `gamedata\\textures` folder
- Variable Resolution


themrdemonized:
The only changed file is .gitignore, and i don't think its intended. Revert changes to .gitignore and check your PR again

antglobes:
Should all be done and uploaded now. Had a little trouble reverting so its not clean, but it should still all be there.

themrdemonized:
Delete this file `src/batch_build - Copy.bat` from PR as well please

PR #396, opened 2025-12-21
Development changes for CProjector, CWeaponStatMgun, bind_monster.script

GhenTuong:
- CProjector: Further development. Export functions and register class to be useable in LUA.
- CWeaponStatMgun: Register class to be useable in LUA. Fix some minor bugs/oversights. (eye_direction, animations, kick owner out on net destroy).
- bind_monster.script: Add an option to make online mutant squads run toward squad target instead of walking. (for making mutant attack quests).


themrdemonized:
`Add a BONE_POSITION_WARNING to allow modders to disable warning spam of bone_position when enabling debug.`

There is already `print_bone_warnings` console command that toggles it, no need for another define

GhenTuong:
BONE_POSITION_WARNING changes was removed.

themrdemonized:
Also i think `S_STMGUN` and `S_SEARCH` should be changed to `_S_STMGUN` and `_S_SEARCH` respectively, that way they will be added in the end of existing classes, so it won't break their order. I honestly don't know how critical it is for those classes, so let me know if it works fine either way

GhenTuong:
It worked fine to me so far but let be safe than sorry.
Changes to add "_" to class names were checked in.

PR #397, opened 2025-12-23
VS2022 Hot Reload & Debug configuration

lulnope:
VS2022 Hot Reload & Debug configuration that allows for quick feedback flow when poking around the engine. See the short video presentation linked in readme for proof of usefulnes: https://youtu.be/MmZwyM2QO38


themrdemonized:
Thanks for PR. I will enable Hot Reload feature for existing Verified configuration, which is one without any optimizations, IMO the new configuration would be redundant for that. I'll take your readme guide and put into separate commit

PR #408, opened 2026-01-27
Fixed spinlock in FPS limiter

leragequit:
Currently it seems that xray does a spinlock when #ECO_RENDERER is set and the game is in menu, pause or framelimited.

Instead of spinlocking we try to sleep first, but leave a small overlap to spin for precision at higher framerates

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -173,8 +173,116 @@ jobs:
         # Duration after which artifact will expire in days. 0 means using default retention. Minimum 1 day. Maximum 90 days unless changed from the repository settings page.
         retention-days: 1
 
+  build_mt:
+    name: build-mt-job
+    runs-on: windows-latest
+
+    strategy:
+      matrix:
+        buildconf: ["DX11", "DX10", "DX9", "DX8", "DX11-AVX", "DX10-AVX", "DX9-AVX", "DX8-AVX"]
+
+    steps:
+    - name: Configure Pagefile
+      # You may pin to the exact commit or the version.
+      # uses: al-cheb/configure-pagefile-action@86589fd789a4de3e62ba628dda2cb10027b66d67
+      uses: al-cheb/configure-pagefile-action@v1.3
+      with:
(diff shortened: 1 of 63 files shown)


erepb:
vs2026 changes probably not needed?

leragequit:
done, reverted that change.

comrade62:
FYI a more efficient fix for this issue (using _mm_pause()) has been implemented in this repo/commit: https://github.com/CnRJay/xray-monolith-gc64/commit/528106d3674a107e0abb20ee33f663edbb61096b, maybe they'd be ok with sharing?

leragequit:
Thanks, i read through it and did some reading up. i hope the current version is fine now :D

themrdemonized:
Sorry, had to revert it

Works inconsistently, doesn't apply from the game launch only after alt tabbing or changing r__framelimit to 0 and back.

I guess open new pull requests when its ready again, accepted too early

leragequit:
Thanks. I will. Oddly enough, would this also apply to the old spin logic as the logic to trigger it remained unchanged?

themrdemonized:
I didn't notice any inconsistencies with old framelimiter, need more testing

leragequit:
Hey, I am still wondering how to reproduce your findings, so i tried the reproduce the issue you had.

So i did some testing, because i had a debugger attached i had to set ps_framelimiter to 1/60 in order to get some results. 

I started the anomaly (with gamma) and while the logo screen was up, i attached a debugger and profiler. so no tabbing in and out. i only stayed in the main menu screen. 
Screenshot 2026-02-04 210306

this is the old code running, with the polling approach. 

Screenshot 2026-02-04 210853

here we have the new code, that sleeps and polls only the tail. 

Could you maybe either take another look or give me some pointers on how to reproduce your findings?

themrdemonized:
i gotta say i was testing it on MT branch since its where most of my attention goes now and didn't test on main branch, so sorry for not mentioning it. If you are interested you might want to take a look how it works there. I'll revisit this later

PR #410, opened 2026-01-29
fix inertia offset movement when using canted aim

PrivatePirate97:
Usually the inertia offset movement direction is always fixed relative to weapon rotation around barrel axis which leads to weird movement when using canted aim (weapon z rotation =/= 0). This edit simply counterrotates the inertia offset position vector to compensate the rotation. Now the offsert inertia movement direction is always the same no matter how much the weapon is rotated around the barrel axis.

Preview:
before scaled 2 after scaled 2


PrivatePirate97:
whoops, accidentally added another commit

PR #416, opened 2026-01-31
actually fix sound device selection and autoswitch

erepb:
i thought i tested with device name that doesn't exist. apparently i did not. this fixes the crash


erepb:
got successful test results from a guy with korean symbols in sound device names

PR #417, opened 2026-02-01
decoupled horz recoil, non-linear inertia movement, additions to lua_help_ex.script

PrivatePirate97:
**1. Decoupled Horizontal Recoil**
Usually horz recoil offset angle is coupled with and therefore dependent on vert recoil offset angle. This causes horz recoil to be practically zero when starting to shoot and slowly increasing with higher vert offset angle. This approach removes this dependency, horz recoil has full effect on first shot. Attention: unlike vert recoil, horz recoil does NOT increase with amout of bullets fire. The magnitude of the offset is random with each shot.
This approach makes more sense imo and also makes gunplay a little more challenging. Can be toggled with the cvar 'g_decouple_horz_recoil', disabled by default. Should this get an option in modded exes settings, e.g. in General/Aim? If yes then I'll add it to the PR.

**2. Non-linear Inertia**
Currently inertia movement uses linear functions to calculate any offsets which makes movement look quiet robotic. This adds non-linear offset calculation functions for offset return movement (after changing view direction) and weapon tilt back movement (after walking sideways) to make the movement feel/look more organic, difference is subtle though. Can be toggled with the cvar 'g_use_non_linear_inertia', disabled by default.

**3. Additions to lua_help_ex.script**
Added new color exports that I forgot to add in my last PR. Also added the cvars from 1. and 2.

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -90,6 +90,16 @@
         int get_modded_exes_version() // Returns modded exes version in integer format
         table get_string_table() // Returns all translated strings in [string id] = string text format 
         player_hud* get_player_hud()
+
+        -- color getters / setters
+        ClrGetA(number argb)
+        ClrGetR(number argb)
+        ClrGetG(number argb)
+        ClrGetB(number argb)
+        ClrSetA(number argb, number a)
+        ClrSetR(number argb, number r)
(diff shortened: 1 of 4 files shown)


themrdemonized:
Thanks, i'll add the options myself

PrivatePirate97:
@themrdemonized alright. :)

PR #418, opened 2026-02-01
various window related fixes

erepb:


erepb:
~couldn't test dx10. both release binaries and built locally crash instantly~

PR #424, opened 2026-02-08
Installation based instance mutex

Tosox:


Tosox:
@themrdemonized Could you check out the changes if you get around it?

themrdemonized:
thanks it works
изображение

Tosox:
Perfect, thank you

PR #426, opened 2026-02-09
fix race in refcount

erepb:
render thread and lua calls (map_add_object_spot for example) can access same refcounters which in case of a race gives crash when counter is decremented more than incremented. 

main change is in xr_resource, two other were changed just in case because they use same pattern.


themrdemonized:
For resources, i tested atomic counter in MT version before, it lead to severe performance issues. I chose critical section blocking instead for resources, might port it later.

str_value does use atomic there

erepb:
i see. feel free to close PR then. warfare mode seems to more extensively use affected lua calls and crashes quite regularly. backport is highly appreciated

PR #427, opened 2026-02-10
Add console execution intercept callback with command consumption

VodoXleb:


Xottab-DUTY:
This functionality isn't good, console should be above the scripts where a modder can do everything, including something bad.
I believe we should not give an ability to intercept console calls.

erepb:
wonder what is the usecase

themrdemonized:
I agree, that can lead to annoying stuff from 3rd party and can spiral out of control

Xottab-DUTY:
> wonder what is the usecase

In the past, old mods used to check some console values and revert them if user has changed the values, e.g. to prevent the user from activating cheats or cheaty commands, to adjust graphics settings during gameplay (e.g. reduce visibility distance when a player entered some area, boost `cam_inert` when player is drunk, etc.)
Since we have engine sources, it's better to implement such gameplay features and restrictions via a dedicated functionality.
E.g. user can change `cam_inert`, but the actual camera inertia value would use a different variable in the code, not tied to the console, thus won't be changed because the player is drunk. In fact, a mod won't need to touch/check `cam_inert` at all then.

It would be good if @VodoXleb would tell us the use case, then we probably could help with an alternative solution.

PR #432, opened 2026-02-14
When the wedge is eliminated, the cartridge disappears from the magazine

knallpsi:


damoldavskiy:
That change should be either under a flag or under a new weapon param. Vanilla uses usual reload anims as misfire anims, also there are misfire animations that don't actually discharge a cartridge, for both those cases this change wouldn't make sense

knallpsi:
> That change should be either under a flag or under a new weapon param. Vanilla uses usual reload anims as misfire anims, also there are misfire animations that don't actually discharge a cartridge, for both those cases this change wouldn't make sense

Thank you for your clarification, I will know :)

Issue #401, opened 2026-01-15
Barrel launcher ammo lost and fire mode reset when mounting/detaching scope [Possible solution inside]


tomilovsenya:
As far as I'm concerned, this happens because _try_to_clone_object_ in **alife_simulator_script.cpp** (traced it all the way from _item_weapon.attach_scope_ function) doesn't also try to clone fire mode and grenades count. There are some mods that try to fix this the script way but I'm pretty sure this issue has to be fixed on engine's level.

Pretty sure the fix should be like this in **alife_simulator_script.cpp**:

	clone->wpn_flags = wpnmag->wpn_flags;
	clone->m_addon_flags = wpnmag->m_addon_flags;
	clone->m_fCondition = wpnmag->m_fCondition;
	clone->ammo_type = wpnmag->ammo_type;
	clone->m_upgrades = wpnmag->m_upgrades;
	clone->set_ammo_elapsed(wpnmag->get_ammo_elapsed());
	// Fix start
	clone->m_iCurFireMode = wpnmag->m_iCurFireMode;
	clone->a_elapsed_grenades.grenades_count = wpnmag->a_elapsed_grenades.grenades_count;
	// Fix end

Might need some additional checks or syntax check but I hope you get the idea. Sadly can't be too bothered to compile it and test myself as well as create a PR. Hopefully it doesn't take more than 5 minutes to implement.

tomilovsenya:
Ok so I tested it and the fire mode part doesn't seem to have any effect and somehow my underbarrel launcher doesn't work so I couldn't test this part. Still, the issue needs to be fixed.

PR #440, opened 2026-02-24
Persist weapon fire mode & grenade launcher data

Tosox:
Related to issue #401

### DIFF
--- a/src/xrGame/WeaponMagazined.cpp
+++ b/src/xrGame/WeaponMagazined.cpp
@@ -1,4 +1,4 @@
-﻿#include "pch_script.h"
+#include "pch_script.h"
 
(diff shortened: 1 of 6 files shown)


damoldavskiy:
Just wanted to say, a new attachment system is being worked on, so at some point scope attaching/detaching won't cause weapon respawning and all script-side logic related to it is going to become obsolete. I'd suggest to wait, or at least limit the fix to sensitive state parameters, like loaded shell in GL.

Tosox:
Thanks for pointing this out, but yes, I am aware that after we get the new attachment system this will be barely relevant anymore. But it will still fix `alife_clone_weapon` calls in the future, when for example addons use this function.

PR #447, opened 2026-02-27
New filesystem

knallpsi:
Most of the file system has been improved and optimized, and the game's startup speed should increase (I hope so..)

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,8 +246,6 @@ Main and MT:
 
 MT:
   * Fixed buggy lighting on objects in DX8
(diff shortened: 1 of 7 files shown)


damoldavskiy:
Would be nice to see a wider description of what's done and why the game startup speed should increase. Even better if there are some actual measurements :)

knallpsi:
> Would be nice to see a wider description of what's done and why the game startup speed should increase. Even better if there are some actual measurements :)

Sorry, it's night, and I physically can't write a detailed description, I'll do it tomorrow afternoon, thank you. 😇

PR #473, opened 2026-03-15
expose graph_reachable to Lua scripts

erepb:
Helpful for lua targeting (e.g., warfare mode) to avoid setting unreachable targets


erepb:
actually, looks like everything is reachable and issue was in another place

PR #475, opened 2026-03-17
Update poltergeist.cpp

leer-h:
A small tweak to ensure the dead poltergeist's body is displayed correctly when saving and loading. This tweak only makes sense if changes have been made to `bind_monster`, since in the base game the poltergeist is removed after death.
Created for this mod(or similar) to prevent situations where the body physically exists but is visually hidden.
https://www.moddb.com/mods/stalker-anomaly/addons/poltergeist-dead-body


StalkerBoss:
Thanks for your work!

PR #508, opened 2026-04-12
CRayPick: implement get_normal

GhenTuong:
Can use get_normal() to get the normal vector of the surface where CRayPick hit. The function only works for static geometry.


themrdemonized:
Add new export into lua_help_ex

PR #531, opened 2026-05-11
Burer and poltergeist telekinesis fix

Orleonn:
Fixed an issue where objects thrown via telekinesis by burers and poltergeists were not tracked properly, making it impossible to register their hits on the actor. Now, the monster that launched the object is correctly passed to SHit as the damage source, instead of the CActor placeholder.


themrdemonized:
Remove actor callbacks from this PR and open new PR with those

Orleonn:
Done. Sorry for the mess.

PR #535, opened 2026-05-17
PDA Map QoL features

Leyths:
**Changes**
- Map names which can take an optional colour and offset (see Cordon in the video attached) drawn at lower zoom levels and disappear as you zoom in closer.
- The map name overlay can be disabled via a settings option (Settings -> Modded EXES -> Visual -> UI/HUD) for players who find it immersion breaking. By default without the label set in the LTX there will not be any names drawn, so this will not change existing behaviour of mods.
- LUA callbacks for when the player clicks on map markers so you can toggle scripts (in the attached video, having the level transitions bring the map to the place the player will be transitioned to - this specific implementation is not a part of this PR). But this could be used for other behaviors. I also exposed the PDA map to LUA so scripts can pan the maps to levels or specific points.

Map names, colours and offsets configured in the game_maps_single.ltx:

```
[l01_escape]
global_rect     = 357.0, 2022.0, 505.0, 2319.0
weathers        = atmosfear
music_tracks    = l01_escape_musics
label           = st_label_l01_escape
label_scale_max = 2.5
label_color     = 255, 80, 80, 255
label_offset_x  = 20
label_offset_y  = 100
```

Demo:

https://github.com/user-attachments/assets/a29a44ae-c30b-4b22-9976-b4dcca610062

(Please ignore the broken looking Agroprom map in the video - this is an incomplete part of a map mod that's being worked on and is not related to this set of changes).

### DIFF
--- a/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -152,6 +152,9 @@
 	<string id="ui_mm_modded_exes_visual_ui_hud_use_english_text_for_missing_translations">
 		<text>Use English Text For Missing Translations (use_english_text_for_missing_translations)</text>
 	</string>
+	<string id="ui_mm_modded_exes_visual_ui_hud_pda_show_map_labels">
+		<text>Show Level Names On Global Map (pda_show_map_labels)</text>
+	</string>
 
 	<!-- Crosshair -->
 	<string id="ui_mm_crosshair_modded_exes">
@@ -680,7 +683,7 @@
 	<string id="ui_mm_modded_exes_control_pda_pda_map_zoom_out_to_mouse">
 		<text>Map Zoom Out Relative To Mouse (pda_map_zoom_out_to_mouse)</text>
 	</string>
-	
+
 	<!-- Gameplay -->
 	<string id="ui_mm_menu_gameplay">
 		<text>Gameplay</text>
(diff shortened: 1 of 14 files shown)


themrdemonized:
Document new script exports into lua_help_ex.script please

Leyths:
@themrdemonized done :)

themrdemonized:
https://github.com/themrdemonized/xray-monolith/pull/535/changes#diff-ec9cb8778410da59ae67b0aba31493390077d0ec8ce21100686b79c7184b65e5R993
I think since you added a lua call for clicking on the map spot, you should add a callback for this so that other mods could use this functionality without changing pda script

Add this into callbacks_gameobject.script
```
AddScriptCallback("on_map_spot_selected")
function pda.map_spot_selected(id, level_name, spot_type)
	SendScriptCallback("on_map_spot_selected", id, level_name, spot_type)
end
```

Leyths:
One question on docs: I'd kept a namespace pda { function map_spot_selected(...) } entry in lua_help_ex.script from the previous feedback. Now that pda.map_spot_selected is an internal bridge (like _G.COnRightClickMap, which isn't in lua_help_ex.script either), and the user-facing API is the on_map_spot_selected script callback documented inline in callbacks_gameobject.script, do you think the lua_help_ex.script entry should be removed for consistency? Happy to drop it if you agree.

themrdemonized:
yea you should remove pda one, since its not engine export. lua help ex contains only engine exports

PR #537, opened 2026-05-17
CTexture: GIF animation support

Orleonn:
Added support for reading gif animations.
Added [giflib](https://sourceforge.net/projects/giflib) for image decoding.
This is used for the UI and simplifies integrating animation playback instead of the texture sequence (.seq).

Known issue:
- ~~animation slows down at low FPS~~


Xottab-DUTY:
I had few hesitancies about using GIF, but after thinking about it, they vanished.
Simplicity is good. Using gif simplifies things, so it's good.

> animation slows down at low FPS

GIF possibly could replace seq usage entirely, if not this bug.
Maybe peek at how avi/ogm/seq is handled to get the clue. E.g. you probably need to align gif frame selection with delta time to fix that.

Orleonn:
@Xottab-DUTY Thanks for the feedback!
avi, ogm and seq rely on CRenderDevice::dwTimeContinual, but since GIF frames have individual frame delays instead of a fixed frame rate, we can't just mathematically jump to a specific frame based on timestamps alone.
I will keep experimenting with it to fix this issue. Thanks again for suggesting the time delta approach!

themrdemonized:
Good job. While its simple to use for already ready files, GIF has limitations for image quality like only 256 colors. Maybe you can discover how to enhance .seq format to support variable frame time, maybe via separate formatting, but its a question for another PR

Orleonn:
@themrdemonized Thanks! Honestly, I don't see any point in upgrading .seq further.
If we want variable frame rates and 12-bit color depth, adding AVIF support would be a much better solution down the road, though it's just a theory for now.

PR #538, opened 2026-05-18
feat(alife): per-actor-level clsid-filtered object iteration to Lua

damiansirbu:
## Summary

Adds `alife():iterate_level_objects_of_clsid(script_clsid, fn)` Lua bind. C++ walks the actor's per-level `CSE_ALifeDynamicObject` set held in `CALifeGraphRegistry`, filters by `m_script_clsid` in the loop, calls the Lua functor only on matches.

Adds `CALifeGraphRegistry::level_exists()` so the bind can guard the pre-actor-spawn window when `m_level` is null.

A script that scans alife objects of a known class on the actor's current level has two options today. Option one walks a Lua-side global registry such as `SIMBOARD.squads`, `SIMBOARD.smarts_by_names`, or `treasure_manager.caches`, and resolves each entry through `alife():object()` and `game_graph()` to read its level. The walk pays per-entry luabind cost across hundreds of entries. On a populated session, around 80% of those entries fail the level check after the resolve. Option two is the existing `iterate_objects` bind, which visits every alife object on every level. A populated session has 5000-15000 entries in that global registry.

The engine already maintains a per-actor-level CSE registry at `CALifeGraphRegistry::level().objects()`, built at actor spawn and kept current by add and remove on object register and unregister. This bind exposes a class-filtered iterator over that registry, pushing both the level filter and the class filter into C++.

## Use cases

Any modder running per-class scans on the actor's level benefits.

- Squad scans by faction or distance on the actor's level via `clsid.online_offline_group_s`. Replaces global `pairs(SIMBOARD.squads)` walks where most entries fail the level check.
- Smart terrain scans via `clsid.smart_terrain`. Avoids walking `SIMBOARD.smarts_by_names` across every level.
- Anomaly scans by type clsid. Avoids iterating `db.anomaly_by_name` across both levels and types.
- Per-class NPC sweeps for telemetry, debug overlays, or AI metrics.

## Performance

On a level with around 1500 dynamic objects and around 100 of the target class:

- The C++ map walk plus the int compare per entry costs 50-150us total.
- Lua functor calls fire only on the matching entries, costing 100-500us depending on the functor body.

Lua-side patterns that walk 700-800 globally registered entries today typically cost 1.5-2.5ms per scan. A class-filtered actor-level iterator drops that to under 1ms on the common case.

## Registry contents

The actor-level registry holds every `CSE_ALifeDynamicObject` on the level. The contents include NPCs, mutants, squads, smart terrains, anomalies, and loose physics objects. A typical populated level holds 1000-3000 such entries. A bind without a class filter would pay one Lua functor call per object before any filtering, landing net-neutral or worse than current Lua loops. The class filter runs in C++ to keep the bind useful.

## Iterator-invalidation constraint

The Lua functor must not release the currently iterated entity. The bind uses range-for over `xr_map`, and `std::map::erase` invalidates the erased iterator. Releasing a non-current entity is fine since `std::map` does not invalidate other iterators on erase. Same constraint as the existing `iterate_objects` bind. Documented in a comment line above the function.

## Cross-level scope

`CALifeLevelRegistry` is per-actor-level only. Cross-level scans keep the existing Lua-side fallback over the global registries. Anomaly level transitions tear down and rebuild `CALifeSimulator` via the save and autoload cycle, so the bind is unreachable from Lua during the transition window.

## Files changed

Three files, 20 lines added in total.

- `src/xrGame/alife_graph_registry.h` declares public `level_exists()` next to `level()`.
- `src/xrGame/alife_graph_registry_inline.h` defines `level_exists()` as `return (!!m_level);`.
- `src/xrGame/alife_simulator_script.cpp` adds the free function `CALifeSimulator__iterate_level_objects_of_clsid` and the matching `.def()` bind next to the existing `iterate_objects`.

## Lua usage

```
alife():iterate_level_objects_of_clsid(clsid.some_class_s, function(se)
    -- inspect se, accumulate, return true to stop early
    return false
end)
```

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -118,6 +118,7 @@
     alife() {
         number object_count()
         void iterate_objects(function(se_obj))
+        void iterate_level_objects_of_clsid(number script_clsid, function(se_obj)) -- walks actor's current-level CSE registry filtered by m_script_clsid; functor returns true to stop early; must not release the currently iterated entity
         void force_update() -- forces an immediate alife scheduled update cycle
     }
 
--- a/src/xrGame/alife_graph_registry.h
+++ b/src/xrGame/alife_graph_registry.h
@@ -75,6 +75,7 @@ public:
 	IC void change(CSE_ALifeDynamicObject* object, GameGraph::_GRAPH_ID game_vertex_id,
 	               GameGraph::_GRAPH_ID next_game_vertex_id);
 	IC CALifeLevelRegistry& level() const;
+	IC bool level_exists() const;
 	IC void set_process_time(const float& process_time);
 	IC CSE_ALifeCreatureActor* actor() const;
 	IC const GRAPH_REGISTRY& objects() const;
--- a/src/xrGame/alife_graph_registry_inline.h
+++ b/src/xrGame/alife_graph_registry_inline.h
@@ -14,6 +14,11 @@ IC CALifeLevelRegistry& CALifeGraphRegistry::level() const
 	return (*m_level);
 }
 
+IC bool CALifeGraphRegistry::level_exists() const
+{
+	return (!!m_level);
+}
+
 IC void CALifeGraphRegistry::change(CSE_ALifeDynamicObject* object, GameGraph::_GRAPH_ID tGraphPointID,
                                     GameGraph::_GRAPH_ID tNextGraphPointID)
 {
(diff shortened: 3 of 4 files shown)


themrdemonized:
Add new export into lua_help_ex.script please

damiansirbu:
Thanks for the review. Manifest entry added in 65ec04e3 — `iterate_level_objects_of_clsid` now declared in the `alife()` block of `lua_help_ex.script`.

PR #541, opened 2026-05-22
Make splash logo window not overlap other windows

AzuNar:
It's irritating when splash screen is on top of other windows and you have to wait for a game to be loaded

### DIFF
--- a/src/xrEngine/x_ray.cpp
+++ b/src/xrEngine/x_ray.cpp
@@ -1050,9 +1050,9 @@ int APIENTRY WinMain_impl(HINSTANCE hInstance,
 	SetWindowPos(
 		logoWindow,
 #ifndef DEBUG
-		HWND_TOPMOST,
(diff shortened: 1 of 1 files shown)


damoldavskiy:
But the splash logo being showed means the game was launched a second ago, doesn't it mean the user wants to see it instead of windows in the background? 🤔

The change is questionable

AzuNar:
> But the splash logo being showed means the game was launched a second ago, doesn't it mean the user wants to see it instead of windows in the background? 🤔
> 
> The change is questionable

Then, make a window appear on top but it can be unfocused, so user can focus another window above logo

damoldavskiy:
> Then, make a window appear on top but it can be unfocused, so user can focus another window above logo

That sounds right to me

themrdemonized:
Implemented separately in 07fb2cedf0e8d1ab9b724094d2abfd339ac5731a

PR #545, opened 2026-05-27
Disabled MSAA commands.

LVutner:
Dead tech.


damoldavskiy:
Isn't it the MSAA that is enabled in Anomaly by default? 😅

LVutner:
yos

LVutner:
This one can be added too, I don't think you maintain MSAA pipeline anymore.

PR #553, opened 2026-05-31
fix busyhands: remove non-savable children

erepb:
nonsavable items became orphaned and reusing their ID lead to busyhands


themrdemonized:
изображение
started to get this type of spam

erepb:
interesting. investigating

erepb:
apparently NPC have 2 bolts with them. one that is alife controlled and another is client only. new PR created

PR #554, opened 2026-05-31
skip member-less squad update

erepb:


themrdemonized:
How it is possible to have a squad with no members

erepb:
offline combat eliminates all members, but actual destroy of the squad is delayed for some time. usually not a problem, it's few? ten-ish? frames, but with force update called every frame and number of objects to update set to max - game tries to update these. 
P.S. not dangerous, just prints invalid id messages when it tries to access commander_id which is 65535

PR #560, opened 2026-06-03
Add g_launcher_dynamic_range_zoom console variable to toggle ballistic aim assist

TheLostInPlace:
Adds a new bool console variable (defaults to true to preserve pre-PR saves) that gates the zoomed ballistic arc aim assist in all CRocketLauncher-derived weapons: WeaponMagazinedWGrenade (the UBGL grenade), WeaponRG6, and WeaponSSRS.

When true the engine redirects the projectile to arc toward the crosshair target via TransferenceAndThrowVelToThrowDir(), matching original behaviour. When false, the projectile fires in the raw aim direction with no correction.

The option is exposed in the Modded Exes 3D Ballistics settings page with a tooltip that clarifies scope (launcher weapons only, hand-thrown grenades and RPG7 are unaffected).

Claude was used to sanitize the code and make sure it doesn't break anything downstream. Testing so far appears to be working.


TheLostInPlace:
[Proof of Concept](https://streamable.com/e/mhtwy2) video of how it would affect the UBGLs, the text is still a little too long so I might just shorten it even more to say "Arc UBGL projectiles to crosshair when zoomed"

TheLostInPlace:
Accidentally pushed a different PR I am working on, should be fixed now

PR #561, opened 2026-06-03
fix portraits stretch

erepb:


themrdemonized:
How does it look like, before and after

erepb:
image
image
I think this is everything affected. I wonder why portraits size doesn't match any of target rectangles aspect ratio.

erepb:
My initial attempt was to preserve aspect ratio and just fit inside. But i think this way it looks better and we don't really loose any important details because portraits are usually centered

PR #562, opened 2026-06-03
floating artefact fix

lulnope:
Artefacts sometimes get stuck floating in grav anomalies. I've narrowed down the cause to a suspicious `power.y += 1` line in `Artefact.cpp`. This PR gates the power adjustment behind a ray query checking if the artefact is close to the ground. This allows the artefact to at least sometimes get within range of actor's hands so that it can be collected as opposed to floating high up in the sky.

Video showcasing the difference in behavior: https://youtu.be/2TQqZqv-qwg

### DIFF
--- a/src/xrGame/Artefact.cpp
+++ b/src/xrGame/Artefact.cpp
@@ -611,8 +611,23 @@ void SArtefactDetectorsSupport::UpdateOnFrame()
 		float cosa = v.dotproduct(dir);
 		if (v.square_magnitude() < (0.7f * 0.7f) || (cosa < cos_et))
 		{
-			Fvector power = dir;
-			power.y += 1.0f;
+            Fvector power = dir;
+
+            {
+                Fvector start = m_parent->Position();
+                Fvector dir_down;
+                dir_down.set(0, -1, 0);
+                const float ground_check_dist = 0.75f;
+
+                collide::rq_result R;
+                if (Level().ObjectSpace.RayPick(start, dir_down, ground_check_dist, collide::rqtBoth, R, m_parent)) {
+                    if (R.range <= ground_check_dist) {
+                        power.y += 1.0f;
+                    }
+                }
+            }
+
+            //power.y += 1.0f;
 			power.mul(m_path_moving_force);
 			m_parent->m_pPhysicsShell->applyGravityAccel(power);
 		}


themrdemonized:
Hm maybe you should remove that +1 on height at all? I kinda don't understand why devs did that in first place

erepb:
that +1 is probably to make it shoot at angle up, like throwing a ball. otherwise it will start moving along straight line to it's destination an gravity will pull it down. i feel this PR is just hiding the problem. original idea probably was to make artifact jump between random points in the list. but that constant +1 seems to be stronger than gravity
or maybe to make it fly via straight line to destination. too late to actually read how that power is used

lulnope:
i think that +1 is there to allow artefacts to jump over obstacles. Removing it entirely will likely result in a lot of artefacts stuck somewhere in a small ditch not having enough strength to get out of that position.

Judging by observable behavior - grav anomailies seem to add their own +y component on top of it so maybe in grav anomalies it won't be a problem but in other types of anomalies the arts would just get stuck.

erepb:
so looks like +1 is needed only in non-grav anomalies?

lulnope:
So I analyzed what's happening a little bit more and here are my findings. Note that I haven't debugged any of this, this is going off of dry code reading:

- in grav anomalies the gravity gets disabled, see this file: https://github.com/themrdemonized/xray-monolith/blob/all-in-one-vs2022-wpo/src/xrGame/NoGravityZone.cpp#L14
- if artefact has a patrol path it follows then it gets the jump-over-the-obstacles behavior that's changed in this PR: https://github.com/themrdemonized/xray-monolith/blob/719b44bd10bd5293a3255d83c8c002d11130964e/src/xrGame/Artefact.cpp#L615
- in grav anomaly fields the artefacts with patrol paths get yeeted up by both of the above mechanisms
- at some point the artefact gets so high that the patrol path logic that pokes the artefact to move results in pokes that point pretty much exactly down. The y component gets close to -1. It never quite gets to -1 but it gets close.
- add to the picture some air resistance and the +1y from jump mechanism gets cancelled out to 0 by the patrol path poke mechanism + air resistance
- result: artefact starts floating high over the ground with little lateral movement.
- result with the proposed fix: the proposed fix disables the +1y effect once the artefact rises a little over the ground which allows the patrol path mechanism to pull the artefact back down into a reachable position since the patrol path pulling mechanism does not have to fight the +1y into an equilibrium.

Summary: I agree with both of you that the fix is ass as fuck but it's slightly less ass than current state of things and it seems to address the issue from my research. Also the fix looks exactly like the jump mechanism in GraviArtifact.cpp so I'm not introducing any novel weirdness in the logic there.

I'm open to suggestions on how to improve the solution or explanations on how my reasoning is wrong.

unrelated note: some artefacts are "grav artefacts" and they get their own dedicated jump logic you can find here: https://github.com/themrdemonized/xray-monolith/blob/719b44bd10bd5293a3255d83c8c002d11130964e/src/xrGame/GraviArtifact.cpp#L52

unrelated note2: https://github.com/themrdemonized/xray-monolith/blob/719b44bd10bd5293a3255d83c8c002d11130964e/src/xrGame/GraviZone.cpp at first glance seems to be the code responsible for another artifact bug where a particular anomaly type tends to yeet artifacts god knows where at light speeds.

lulnope:
ix-ray has the exact same code in NoGravityZone.cpp and Artefact.cpp so it seems to suffer from the same issue? Hard to tell as commits are described in russian and i don't understand that language

https://github.com/ixray-team/ixray-1.6-stcop/blob/514e0d89b971e6ae95915c1ec00c1a9142ef1c21/src/xrGame/NoGravityZone.cpp

https://github.com/ixray-team/ixray-1.6-stcop/blob/default/src/xrGame/Artefact.cpp#L657

a quick look through log histort of ix-ray also shows no commits relating to floating artefact fixes (i just grepped for "artefact") so they don't have a better idea either

themrdemonized:
Neither ixray nor openxray addressed this issue, so no point in looking at them. For now this PR is good enough to reduce the issue

PR #563, opened 2026-06-03
feat(ai): Lua Combat AI Hook System

TheLostInPlace:
﻿# Lua Combat AI Hook System

## What this adds

A set of Lua hooks, NPC object exports, and console variables for combat AI behavior that were previously hardcoded with no script access. Everything defaults to the original engine values - no behavior changes for existing saves or mods that don't use the new hooks.

**Why these specific gaps:** NPCs shot in the open had no code path to seek cover (the existing danger system only fires when `brain().affect_cover()` is already true). Grenade throwing had no Lua gate at all. Enemy acquisition/loss required polling `npc:best_enemy()` every frame. Cover selection distances, danger urgency, and suppression timing were all `static const` values that required a recompile to change.

---

### `ai/stalker/ai_stalker_fire.cpp`

Four new hooks on existing functions:

- **`npc_on_critically_wounded(npc, hit, bone_id)`** - fires once when NPC first goes critical. Placed after rank immunity and bone armor are applied, before `BeforeHitCallback`. Pipeline position confirmed against source.
- **`npc_on_weapon_shot_start(npc, weapon)` / `npc_on_weapon_shot_stop(npc)`** - hooks on the empty `on_weapon_shot_start/stop` virtual stubs. Used together to track burst duration.
- **`npc_on_missile_throw_force(npc, base_force, flags)`** - return-value hook in `missile_throw_force()`. Return value validated (must be `> 0` and `_valid()`), bad returns log with NPC name and fall back to engine default.
- **`npc_on_enemy_wounded_or_killed(npc, victim)`** - fires after kill-cry sound. Added null check on victim before `lua_game_object()`.

Four `static const` values converted to tunable globals:

| Variable | Default | Range | Effect |
|---|---|---|---|
| `ai_fire_make_sense_interval` | 10000 ms | 0-20000 | How long NPCs fire at last known position after losing sight |
| `ai_fire_range_extension` | 2.5 m | 0-10 | Extra range past pick distance where suppression still makes sense |
| `ai_fire_max_height_diff` | 2.0 m | 0-8 | Max vertical gap between NPC and enemy for firing to make sense |
| `ai_fire_min_dist` | 2.5 m | 0-20 | Minimum pick distance required for firing to make sense |

---

### `ai/stalker/ai_stalker_cover.cpp`

Two new hooks:

- **`npc_on_best_cover_changed(npc, cover_pos, is_smart_cover, smart_cover_name)`** - fires in `on_best_cover_changed()` after internal delegates. `cover_pos` is nil when NPC loses cover. `smart_cover_name` is the level object `cName()`.
- **`npc_on_get_min_combat_dist` / `npc_on_get_max_combat_dist`** - return-value hooks in `compute_enemy_distances()` after weapon-type defaults are set. Override the min/max distance from enemy used when selecting cover. Return values validated, bad values log and fall back. The `goto` for the no-weapon case replaced with `if/else`.

Six hardcoded values converted to globals. Minimums are engine-aware: `3.0m = MIN_SUITABLE_ENEMY_DISTANCE` - below this the cover evaluator gets a zero-width range and always fails.

| Variable | Default | Range | Effect |
|---|---|---|---|
| `ai_cover_search_near_radius` | 10.0 m | 1-30 | First-pass cover search radius |
| `ai_cover_search_far_radius` | 30.0 m | 1-80 | Fallback cover search radius |
| `ai_cover_pistol_max_dist` | 10.0 m | 3-50 | Max cover distance, pistol wielders |
| `ai_cover_shotgun_max_dist` | 5.0 m | 3-20 | Max cover distance, shotgun wielders |
| `ai_cover_sniper_min_dist` | 20.0 m | 0-80 | Min cover distance, sniper wielders (0 = no minimum, any cover distance accepted) |
| `ai_cover_default_max_dist` | 20.0 m | 3-50 | Max cover distance, all other weapons |

---

### `ai/stalker/ai_stalker_misc.cpp`

- **`npc_on_member_death_reaction(npc, member, is_alive)`** - fires in `react_on_member_death()` after the sound plays. Rate-limited in C++ by `TOLLS_INTERVAL = 2000ms` before the hook fires. Added `script_game_object.h` and `ai_space.h` includes.

---

### `stalker_combat_planner.cpp`

- **`npc_on_combat_action_changed(npc, old_op, new_op)`** - fires in `update()` on GOAP action transitions. `old_op`/`new_op` match `stalker_ids.action_*`. Guarded against first-initialization frame (would otherwise pass `u32(-1)` as `old_op`).

---

### `enemy_manager.cpp`

- **`npc_on_enemy_selected(npc, new_enemy)`** - fires in `try_change_enemy()` when `selected()` changes. `new_enemy` is nil on target loss. Replaces per-frame `npc:best_enemy()` polling.

---

### `stalker_property_evaluators.cpp`

- **`npc_on_should_throw(npc, flags)`** - gate hook in `CStalkerPropertyEvaluatorThrowGrenade::evaluate()`. Hooked in the evaluator so blocking prevents `eWorldPropertyShouldThrowGrenade` from being set - no stale world properties. In-flight throws (grenade slot already active) bypass this hook.

---

### `stalker_combat_actions.cpp` / `stalker_danger_unknown_actions.cpp`

Seven more hardcoded values converted:

| Variable | Default | Range | Effect |
|---|---|---|---|
| `ai_close_move_distance` | 1.5 m | 0-6 | Distance at which TakeCover switches from running to in-place |
| `ai_crouch_look_out_delta` | 5000 ms | 0-12000 | LookOut crouch/stand randomization interval |
| `ai_wait_in_smart_cover_time` | 30000 ms | 0-120000 | How long NPC waits at smart cover before giving up |
| `ai_cover_detour_radius` | 5.0 m | 0-15 | Danger zone radius when NPC flanks |
| `ai_cover_detour_time` | 120000 ms | 0-300000 | Duration of that danger flag |
| `ai_cover_unknown_radius` | 5.0 m | 0-15 | Danger zone radius on unknown threats |
| `ai_cover_unknown_time` | 120000 ms | 0-300000 | Duration of that danger flag |

---

### `danger_manager.cpp`

The eight hardcoded danger scores in `do_evaluate()` converted from raw floats to multipliers. The original values (3000, 2500, etc.) stay hardcoded as base scores - the multiplier is applied on top: `result += base_score * multiplier`. At `1.0` the behavior is identical to original. `0.0` disables that danger type.

| Variable | Default | Range | Base score |
|---|---|---|---|
| `ai_danger_ricochet_mult` | 1.0 | 0.0-5.0 | 3000 |
| `ai_danger_attack_sound_mult` | 1.0 | 0.0-5.0 | 2500 |
| `ai_danger_entity_attacked_mult` | 1.0 | 0.0-5.0 | 2000 |
| `ai_danger_entity_death_mult` | 1.0 | 0.0-5.0 | 3000 |
| `ai_danger_corpse_mult` | 1.0 | 0.0-5.0 | 2250 |
| `ai_danger_attacked_mult` | 1.0 | 0.0-5.0 | 2000 |
| `ai_danger_grenade_mult` | 1.0 | 0.0-5.0 | 1000 |
| `ai_danger_enemy_sound_mult` | 1.0 | 0.0-5.0 | 1000 |

---

### `script_game_object3.cpp` + `.h` + script exports

Five new NPC methods. All log `[section_name]` to `xray_*.log` if called on a non-stalker and return `""` safely.

- `npc:affect_cover()` - returns `brain().affect_cover()` - whether the NPC is currently occupying a covered position. Set to `true` by combat actions when the NPC has physically arrived at cover; reset to `false` on every action transition. Use with `BeforeHitCallback` to react when hit in the open. Note: this is not the same as having a cover destination selected - use `npc_on_best_cover_changed` for that.
- `npc:rank_name()` - returns rank id string from `CharacterInfo` (`"novice"`, `"experienced"`, `"veteran"`, `"master"`, `"expert"`). More reliable than `npc:rank()` since raw integer thresholds change per modpack.
- `npc:best_cover_invalidate()` - forces cover re-evaluation next tick. Was internal-only, missing from Lua API.
- `npc:get_current_smart_cover_name()` - level object `cName()` of the occupied smart cover, `""` if none.
- `npc:get_current_loophole_id()` - loophole id within the smart cover, `""` if none.

---

### `callbacks_gameobject.script`

11 new `AddScriptCallback` registrations. All `_G.*` globals wired to `SendScriptCallback` following existing conventions. Return-value hooks use the `flags.ret_value` pattern so multiple mods can register without stomping each other.

---

## Error handling

All new exports and return-value hooks log to `xray_*.log`:
- Non-stalker calls log `"CGameObject: [section] called on non-stalker"` matching the existing `cNameSect_str()` pattern
- Bad return values from the distance and throw force hooks log the bad value and NPC `cName()`, then fall back to engine default

---

## Save compatibility

All defaults match the original hardcoded values exactly. None of the new console vars will be in existing `user.ltx` files, so C++ defaults apply - identical to original behavior. Zero behavior change for existing saves or mods that don't register any of the new callbacks.

---

<details>
<summary>Full hook and console variable reference</summary>

### Hook quick-reference

| Callback | Signature | Fires in |
|---|---|---|
| `npc_on_enemy_selected` | `(npc, new_enemy)` | `CEnemyManager::try_change_enemy()` |
| `npc_on_combat_action_changed` | `(npc, old_op, new_op)` | `CStalkerCombatPlanner::update()` |
| `npc_on_critically_wounded` | `(npc, hit, bone_id)` | `CAI_Stalker::Hit()` on first critical transition |
| `npc_on_best_cover_changed` | `(npc, cover_pos, is_smart, cover_name)` | `on_best_cover_changed()` |
| `npc_on_get_min_combat_dist` | `(npc, base_min, flags)` | `compute_enemy_distances()` |
| `npc_on_get_max_combat_dist` | `(npc, base_max, flags)` | `compute_enemy_distances()` |
| `npc_on_take_cover_destination` | `(npc, cover_pos, enemy)` | `CStalkerActionTakeCover::execute()` |
| `npc_on_danger_location_add` | `(npc, pos, radius, in_open, cover_threatened)` | `on_danger_location_add()` |
| `npc_on_weapon_shot_start` | `(npc, weapon)` | `on_weapon_shot_start()` |
| `npc_on_weapon_shot_stop` | `(npc)` | `on_weapon_shot_stop()` |
| `npc_on_should_throw` | `(npc, flags)` gate | `CStalkerPropertyEvaluatorThrowGrenade::evaluate()` |
| `npc_on_missile_throw_force` | `(npc, base_force, flags)` return-value | `missile_throw_force()` |
| `npc_on_enemy_wounded_or_killed` | `(npc, victim)` | `on_enemy_wounded_or_killed()` |
| `npc_on_member_death_reaction` | `(npc, member, is_alive)` | `react_on_member_death()` |

### `npc_on_danger_location_add` - performance note

Fires for every danger type for every NPC in range. In a full firefight this is 50-100+ calls/sec. Callbacks here must use per-NPC timestamp throttling.

```lua
local npc_danger_next = {}
RegisterScriptCallback("npc_on_danger_location_add", function(npc, pos, radius, in_open, cover_threatened)
    if not (in_open or cover_threatened) then return end
    local tg, id = time_global(), npc:id()
    if (npc_danger_next[id] or 0) > tg then return end
    npc_danger_next[id] = tg + 1000
    npc:best_cover_invalidate()
end)
RegisterScriptCallback("game_object_on_net_destroy", function(obj) npc_danger_next[obj:id()] = nil end)
```

### Pattern A vs Pattern B

**Pattern A** (`_g` lowercase) - engine calls a Lua global directly and uses the return value. Last mod to assign the global wins - two mods assigning `_g.CAI_Stalker__GetWeaponAccuracy` will silently stomp each other.

**Pattern B** (`_G` uppercase + `SendScriptCallback`) - engine calls a global that fans out to all registered listeners. Multiple mods can `RegisterScriptCallback` on the same event without conflict. New return-value hooks in this PR use `flags.ret_value` via `SendScriptCallback` so all listeners cooperate.

### Composite example

```lua
local last_inv, shot_start = {}, {}
local function try_invalidate(npc)
    local tg, id = time_global(), npc:id()
    if (last_inv[id] or 0) > tg then return end
    last_inv[id] = tg + 1000
    npc:best_cover_invalidate()
end

_G.CAI_Stalker__BeforeHitCallback = function(npc, hit, bone_id)
    if not npc:affect_cover() then
        local r = npc:rank_name()
        if r == "novice" or r == "experienced" then try_invalidate(npc) end
    end
    return true
end

RegisterScriptCallback("npc_on_member_death_reaction", function(npc, member, is_alive)
    if is_alive then return end
    local r = npc:rank_name()
    if r == "novice" or r == "experienced" then try_invalidate(npc) end
end)

RegisterScriptCallback("npc_on_weapon_shot_start", function(npc) shot_start[npc:id()] = time_global() end)
RegisterScriptCallback("npc_on_weapon_shot_stop", function(npc)
    local id = npc:id()
    if npc:rank_name() == "novice" and time_global() - (shot_start[id] or 0) > 1500 then
        try_invalidate(npc)
    end
    shot_start[id] = nil
end)

RegisterScriptCallback("npc_on_should_throw", function(npc, flags)
    local r = npc:rank_name()
    if r == "novice" then flags.ret_value = false
    elseif r == "experienced" then flags.ret_value = math.random() > 0.5 end
end)

RegisterScriptCallback("npc_on_missile_throw_force", function(npc, base, flags)
    local m = { novice=0.6, experienced=0.8, veteran=0.9, master=1.0, expert=1.1 }
    flags.ret_value = base * (m[npc:rank_name()] or 1.0)
end)

RegisterScriptCallback("game_object_on_net_destroy", function(obj)
    last_inv[obj:id()] = nil; shot_start[obj:id()] = nil
end)
```

</details>


themrdemonized:
`npc_on_hit_reaction(npc, hit, bone_id, is_in_cover) - fires after inherited::Hit() on living NPCs only. is_in_cover reflects brain().affect_cover() at time of hit.`

Why you can't use `npc_on_hit_callback`? Maybe you can export `brain().affect_cover()`. CStalkerPlanner already has exports, albeit named weirdly

TheLostInPlace:
Ah that's a good catch, I didn't notice because of the odd naming, which brings us to two options:


**Option 1 - Export `affect_cover()`, drop `npc_on_hit_reaction`**

Export `brain().affect_cover()` as `npc:affect_cover()` and use the existing `BeforeHitCallback`:

```lua
_G.CAI_Stalker__BeforeHitCallback = function(npc, hit, bone_id)
    if not npc:affect_cover() then
        npc:best_cover_invalidate()
    end
    return true
end
```

Smaller surface, reuses what's already there. The one difference is `BeforeHitCallback` fires before outfit armor is applied, so `hit.power` isn't fully reduced yet, but for cover reaction purposes that doesn't matter.


**Option 2 - Export `affect_cover()`, keep `npc_on_hit_reaction`**

Export `npc:affect_cover()` either way (useful on its own), and keep `npc_on_hit_reaction` as a separate post-damage notification that fires after `inherited::Hit()` with all armor applied. Keeps cancellation and reaction as separate hooks.

```lua
RegisterScriptCallback("npc_on_hit_reaction", function(npc, hit, bone_id, is_in_cover)
    if not is_in_cover then
        npc:best_cover_invalidate()
    end
end)
```

Either way would work with the "Smarter AI" mod I have in mind and working on paper, which would you prefer?

themrdemonized:
Proceed with Option 1 - Export affect_cover(), drop npc_on_hit_reaction

TheLostInPlace:
Per feedback, changed `npc_on_hit_reaction` with `npc:affect_cover()`

## What changed

### `src/xrGame/ai/stalker/ai_stalker_fire.cpp`

Removed the `_G.CAI_Stalker__OnHitReaction` block that fired after `inherited::Hit()`. The existing `BeforeHitCallback` (`_G.CAI_Stalker__BeforeHitCallback`) already covers this use case — callers can now check `npc:affect_cover()` inside it instead.

### `src/xrGame/script_game_object.h`

Added declaration:

```cpp
bool affect_cover() const;
```

### `src/xrGame/script_game_object3.cpp`

Added implementation of `CScriptGameObject::affect_cover()` — calls `stalker->brain().affect_cover()` with the standard non-stalker guard that logs to `xray_*.log` and returns `false` safely.

### `src/xrGame/script_game_object_script3.cpp`

Added two `.def()` Lua bindings alongside the existing `best_cover_invalidate` registration. `rank_name`, `get_current_smart_cover_name`, and `get_current_loophole_id` are already registered in `script_game_object_script2.cpp` from an earlier commit and were not duplicated here:

```cpp
.def("best_cover_invalidate", SAFE_WRAP(&CScriptGameObject::best_cover_invalidate))
.def("affect_cover",          SAFE_WRAP(&CScriptGameObject::affect_cover))
```

### `gamedata/scripts/callbacks_gameobject.script`

Removed `AddScriptCallback("npc_on_hit_reaction")` and `_G.CAI_Stalker__OnHitReaction`.

---

## Usage after this change

Instead of registering `npc_on_hit_reaction`, use the existing `BeforeHitCallback` and check `npc:affect_cover()`:

```lua
_G.CAI_Stalker__BeforeHitCallback = function(npc, hit, bone_id)
    if not npc:affect_cover() then
        -- NPC was hit in the open, force cover re-evaluation
        npc:best_cover_invalidate()
    end
    return true
end
```

`npc:affect_cover()` returns `true` when the NPC has a cover point selected (`brain().affect_cover()`), `false` when it is in the open.

---

## PR description updated

- Hook count: 5 → 4 in `ai_stalker_fire.cpp`
- NPC methods: 4 → 5 in `script_game_object3.cpp` (added `npc:affect_cover()`)
- Callback registrations: 12 → 11 in `callbacks_gameobject.script`
- Hook quick-reference table: `npc_on_hit_reaction` row removed
- Composite example: updated to use `BeforeHitCallback` + `npc:affect_cover()`

themrdemonized:
Add new script exports into lua_help_ex file (affect_cover, best_cover_invalidate)

TheLostInPlace:
Added the exports and ran a code-review action using Claude to identify any errors I've made, it's caught some optimization points but nothing major so far. Code also compiles and the variables write correctly/read correctly as well.

porkmanager:
So good! Thx for the great contribution!

PR #564, opened 2026-06-08
Detail textures: add v4 format (14-bit ids, up to 16383 objects), read v3+v4

Leyths:
The detail-object limit was 63/level, fixed by the 6-bit id fields in the on-disk DetailSlot. Add a v4 slot (14-bit ids, 20 bytes, fields reordered to pack into 3x u32, ID_Empty=0x3FFF) alongside the legacy v3 (6-bit, 16 bytes).

- DetailFormat.h: DetailSlot_v3 (legacy I/O), wide DetailSlot (working/v4), expand_v3/pack_v3, static_asserts on the 16/20-byte layout, r_id/w_id -> u16.
- DetailManager: read both v3 and v4 level.details; slots are copied into a heap-owned wide array (v3 expanded on load) instead of aliasing the VFS, freed in Unload. Object pool is now xr_vector; dm_max_objects = 16383 with a load-time guard.

Existing v3 levels load unchanged.

I have a fork of the IXray SDK with support for the v4 format, as well as a version of the Skyloader details compiler to pull through these changes to the compiled maps.

----

This allows for more creative use of detail objects during the level build process (more grass types, flowers, stones, other small detail objects that can be scattered around the map).

### DIFF
--- a/src/Layers/xrRender/DetailFormat.h
+++ b/src/Layers/xrRender/DetailFormat.h
@@ -4,57 +4,26 @@
 #define _DETAIL_FORMAT_H_
 #pragma pack(push,1)
 
-#define DETAIL_VERSION		3
+// Detail format: v3 (legacy, 6-bit ids, 16B slot) + v4 (14-bit ids, 20B slot).
+// Layout is shared byte-for-byte across engine, SDK and the details compiler.
+#define DETAIL_VERSION_3	3			// legacy: 6-bit ids, 16-byte slot
+#define DETAIL_VERSION_4	4			// new:    14-bit ids, 20-byte slot
+#define DETAIL_VERSION		DETAIL_VERSION_4	// default for newly written files
+
 #define DETAIL_SLOT_SIZE	2.f
 #define DETAIL_SLOT_SIZE_2	DETAIL_SLOT_SIZE*0.5f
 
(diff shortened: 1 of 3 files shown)


themrdemonized:
Is this crash related to this PR? I attached images, mdmp and log file, the exe is the latest MT release
изображение
изображение
[Downloads.zip](https://github.com/user-attachments/files/29129794/Downloads.zip)

Leyths:
@themrdemonized thanks for taking the time to look at this and post the dump.

I might be reading this wrong, but the version line in the log shows Modded Exes MT-TEST version 2026.06.01, and I think __DATE__ is set at compile time so does that mean the exe was built around 1 June, before the v4 change landed on the 12th? If so the new code wouldn't actually be in the binary that crashed. Worth double-checking against whatever build you've got though, in case I'm misreading the version string.

The crash itself looks like it's in the detail render path but on a shader constant lookup (get_ConstantDirect returning null for array), which I don't think v4 affects but you'd know that area much better than me. Could it be a stale shaders_cache, or something specific to that user's modded detail content? Happy to take another look, let me know what you reckon.

themrdemonized:
That is actually a good cue, the log refers to 06.01 version, but dump is for 06.13, and since both of those point to same error, it is not related to your changes, sorry for confusion

PR #567, opened 2026-06-09
Detail grass instancing, batching and improved detail falloff curve

samloeschen:
Hi, I've implemented some improvements to grass instancing and specifically draw call batching that should provide some significant performance benefits.

in order of significance:
- Grass draws are now batched, requiring no new pipeline state changes or buffer maps between grass draws. Even without instancing shaders, this should provide a significant CPU performance uplift and minor GPU performance uplift
- Grass draw calls are now instanced. I noticed the other feature branches that have this on them, but implemented it separately so that it would work with the batching. Due to the batching, this is mostly a minor CPU performance uplift.
- I've tried a custom falloff curve for grass which in my opinion looks a lot nicer than the default one. Grass is densely clustered around the camera, and falls off with a power curve after a certain distance. This is tunable with the parameters:

- `r__detail_density_knee`: where along 0 -> r__detail_radius the grass begins to drop off
- `r__detail_density_curve`: the shape of the power curve. >1 will have a gentle tail, <1 will have a sharp cliff ==1 will be linear.

currently, adjusting these parameters requires a `vid_restart` since it performs distribution at the spawn level, instead of the culling/submit level.

This is my first time modding xray itself so let me know if there's anything missing. I have found a mod-based workflow to work well for A/B testing, so I've included my modified exe + the necessary shaders to make it work as a zipped mod: [Grass Instancing Fix.zip](https://github.com/user-attachments/files/28773984/Grass.Instancing.Fix.zip)

quick example image from my testing. this is a 10-15 fps boost on my system (9950x3d, 3090) in this area. Less geometry is drawn overall but the visual quality is about the same if not better. 
image


samloeschen:
alrighty, found a few more things to optimize. Big one is that grass instance lists are filled 3x per frame, but if we do fading in the vertex shader instead of on the cpu, we only need to fill once per frame. This also lets us further optimize the draw calls a bit. Two, is that the grass was previously unsorted. It's now sorted by depth (front to back) which really helps with preventing overdraw. Finally, I've compressed the grass instances into 64b, whereas they were 80b previously. 

The same angle that was getting 65 FPS in my previous test now gets 82:

image


Here is a zip containing updated .exes and the grass shader mod you'll need to install through MO2. Be sure to just replace the exes in your bin folder, installing them as a mod doesn't seem to work.

[grass optimizations.zip](https://github.com/user-attachments/files/28820904/grass.optimizations.zip)

TheLostInPlace:
Great optimizations. I ran this locally on my modified engine and it worked beautifully. Haven't noticed any visual artifacting or issues yet.

Priler:
looks dope, although I wonder if this can be implemented with MT branch as well, cuz it contains tons of rendering changes

erepb:
YES! finally biggest FPS killer gets some love

samloeschen:
MT version here, let me know if you want a separate PR: https://github.com/samloeschen/xray-monolith/tree/detail-grass-instancing-mt

themrdemonized:
Yea please make separate PR for MT branch, would be easier for me this way

> MT version here, let me know if you want a separate PR: https://github.com/samloeschen/xray-monolith/tree/detail-grass-instancing-mt

themrdemonized:
For merging into repo you need to do two things:
1. Add changed vanilla shaders for grass instancing
2. For shaders you need to make them compatible with DX10, since DX10 has no instancing. You can refer to grass-instancing-2 branch to see how it was done there

samloeschen:
Alright, I used AI for this one [(commit here)](https://github.com/themrdemonized/xray-monolith/pull/567/commits/6ab86a8c47f44ec2d47f091aad8e4fe005bc7339) because I am unable to get the dx10 version of the exes to run in my environment. I had it follow the grass-instancing-2 branch you suggested. I've reviewed the code and it looks sane, but please smoke test.

updated shaders mod as well:
[grass shaders.zip](https://github.com/user-attachments/files/29117714/grass.shaders.zip)

samloeschen:
corrected shaders:
[grass shaders.zip](https://github.com/user-attachments/files/29154934/grass.shaders.zip)

I missed a typo Claude inserted when I was doing the dx10 stuff 🤦‍♂️

PR #575, opened 2026-06-15
infinite/super long bolts throwing issue workaround

Priler:
When anomalies interaction enabled, prevents NPCs to throw bolts indefinitely.
Without this, they get stuck there. With this change, it takes them few seconds to stop throwing the bolts and proceed whatever they were doing.
Tested in GAMMA 0.9.5 with N.D.A. (NPCs Die in Anomalies) mod.

~~Should also prevent bolts throwing if NPC doesn't have em in inventory.~~


ravenascendant:
because bolts clear themselves from inventories without a script similar to the one keeping track of the players bolts NPCs will have 0 bolts in their inventory after save load making this a complicated way to disable NPCs ever throwing bolts.

Priler:
also exposed `inside_anomaly()` to Lua so we can actually check it without dumb Lua workarounds

Priler:
> because bolts clear themselves from inventories without a script similar to the one keeping track of the players bolts NPCs will have 0 bolts in their inventory after save load making this a complicated way to disable NPCs ever throwing bolts.

fkn Xray

PR #585, opened 2026-06-26
feat: PDA context menu from multiple spots

erepb:
image
image


erepb:
don't know how to move to draft. don't merge. it might have some issues on the scripts side

erepb:
apparently i just forgot to put updated script file into the anomaly installation. all good

PR #588, opened 2026-06-28
feat(renderer): texture eviction system — level transition cleanup and mid-session LRU eviction

noisethanks:
While profiling my playthrough, I noticed the game's memory footprint grew indefinitely until it overflowed VRAM and began filling GTT space, eventually causing a sharp FPS collapse. Profiling with MangoHud and /proc/PID/fdinfo on AMD RDNA3 confirmed GTT growing to 15GB+ mid-session, with VRAM peaking at 11.5GB before FPS collapsed. System RAM ballooned to 23GB and never recovered between sessions.

Investigation revealed that CResourceManager::Evict() is compiled out entirely on DX11/DX10 builds — the body is wrapped in #if !defined(USE_DX10) && !defined(USE_DX11), leaving a complete no-op. As a result, textures are never unloaded during a session, and level transitions do nothing to reclaim GPU memory.

This PR adds two new functions:
EvictAllTextures() — unconditionally releases all non-essential texture GPU resources. Called from CRender::level_Unload(), which is the common teardown path for normal level transitions, fast travel, and main menu exit. Confirmed working: Jupiter → Radar transition drops VRAM from ~11.6GB to ~7.4GB and GTT from ~15GB to ~2.2GB before the new zone begins loading.

EvictStalledTextures() — runs every 600 frames, scans 50 textures per pass, and evicts textures that haven't been used in 1800 frames (~30 seconds at 60fps). On my configuration this frees 10-20MB per pass with no noticeable frame impact.
This is a intentionally conservative first implementation of mid-session texture eviction. A more complete implementation would prioritize high-impact textures by size, leverage existing engine events for targeted eviction (emission end, zone clear, weather transitions), and more aggressively address the shader element ref retention that currently prevents ~60-70% of stale textures from being evicted. Feedback welcome on all of the above.

### DIFF
--- a/src/Layers/xrRender/ResourceManager.cpp
+++ b/src/Layers/xrRender/ResourceManager.cpp
@@ -508,14 +508,25 @@ void CResourceManager::EvictStalledTextures(u32 max_age_frames) {
   u32 scanned = 0;
   for (; I != m_textures.end() && scanned < batch_size; ++I, ++scanned) {
     CTexture *tex = I->second;
-
-    // $user$ textures (render targets, user-managed) already carry bUser=true
     if (tex->flags.bUser) {
       skip_user++;
       continue;
     }
+    if (tex->cName.size() && strstr(tex->cName.c_str(), "$user$")) {
+      skip_user++;
+      continue;
+    }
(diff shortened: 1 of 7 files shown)


themrdemonized:
see #589

PR #589, opened 2026-06-28
feat(renderer): texture eviction system — MT port of #588

noisethanks:
MT port of #588  — see that PR for full context, design rationale, and before/after profiling data.
This PR applies the same texture eviction system to the MT build with the following adaptations:
Threading considerations:

dwLastUsedFrame stamps in Apply() and apply_normal() are placed strictly after the bLoading spin-wait, ensuring we never stamp a frame on a texture that hasn't finished loading on the resource thread
EvictStalledTextures() holds creationGuard for the duration of the scan, preventing the resource thread from creating or deleting textures while the eviction pass iterates m_textures
fastdelegate replaced with xr_delegate throughout to match MT branch conventions

Functionality is identical to the ST PR:

EvictAllTextures() called from CRender::level_Unload() for transition cleanup
EvictStalledTextures() wired into CRender::Render() every 600 frames
Same exclusions: $user$ render targets, $null, UI textures
Same confirmed result: Jupiter → Radar drops VRAM ~11.6GB → ~7.4GB and GTT ~15GB → ~2.2GB on transition

Known limitations are the same as the ST PR — ref retention wall and cold load footprint on large zones remain open problems.

### DIFF
--- a/src/Layers/xrRender/ResourceManager.cpp
+++ b/src/Layers/xrRender/ResourceManager.cpp
@@ -536,6 +536,129 @@ void CResourceManager::Evict()
 #endif	//	USE_DX10
 }
 
+void CResourceManager::EvictStalledTextures(u32 max_age_frames) {
+#if defined(USE_DX11)
+  Msg("* [TexEvict] pass called, frame=%u", RDEVICE.dwFrame);
+  if (!RDEVICE.b_is_Ready)
+    return;
+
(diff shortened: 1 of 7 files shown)


themrdemonized:
The code is effectively the same as already made `UnloadAllTexturesOnLevelUnload` that fires with `clear_resources_on_unload` cvar. Only filtering is slightly different. I'll look at it if its actually better.

PR #593, opened 2026-07-02
Fix nil clsid crash in IsWeapon and IsLauncher

valn-ilyo:
IsWeapon and IsLauncher both build their lookup table with a raw table constructor:

    { [clsid.wpn_ssrs_s] = true }

If clsid.wpn_ssrs_s is nil at the time this runs (for example when a mod or exe build hasn't registered that weapon class yet in the current load order), Lua throws "table index is nil" and the game crashes. This reproduced on every fresh game start when the actor was given a starting item, right after DRX finished setting up the storyline task.

Added a build_clsid_set helper that filters out nil values before inserting them into the table, and switched both IsWeapon and IsLauncher to use it. Behavior is unchanged when every clsid is present. If one is missing, it's now skipped instead of crashing, and any future weapon class added to either list is protected the same way, not just wpn_ssrs_s.

### DIFF
--- a/gamedata/scripts/_g_patches.script
+++ b/gamedata/scripts/_g_patches.script
@@ -681,6 +681,21 @@ _G.alife_clone_weapon = function(se_obj, section, parent_id)
 end
 
 -- Patch _G functions to recognize new weapon classes
+-- Builds a clsid lookup set from a list of clsid values, silently skipping any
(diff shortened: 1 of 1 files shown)


themrdemonized:
It can't be nil because the class is added in the engine, closing this as unnecessary
https://github.com/themrdemonized/xray-monolith/pull/128

PR #599, opened 2026-07-09
Add new Cvar pseudogiant_dodge_stomp_while_falling

Verdatim25:
Adds new Cvar "pseudogiant_dodge_stomp_while_falling"
Adds one more check to prevent pseudogiant stomp attack from landing if actor is in fall state and Cvar is enabled
The Cvar is off by default.
I dont know russian so russian translation was not by me, it was instead by Google Translate. I'd appreciate if someone checked if it is sensible, thanks!


ProTriforcer:
"Уклонение от хлопка псевдогиганта в падении" would be a bit more precise since pseudogiants have more than one attack

Verdatim25:
> "Уклонение от хлопка псевдогиганта в падении" would be a bit more precise since pseudogiants have more than one attack

alright, changed the text to the one you suggested, thanks!

PR #594, opened 2026-07-04
feat(ai): per-NPC aim params and vision speed setters

damiansirbu:
## Summary

Two per-object setters so scripts can shape an individual NPC's reaction curve without touching the global cvars:

- `npc:set_aim_params(max_angle, min_angle, min_speed, predict_time)` — 4 floats on `CAI_Stalker`, consumed by `CSightManager::Exec_Look` (body/head swing speed and firing cone, `sight_manager.cpp:162-178`) and `CSightAction::predict_object_position` (target lead, `sight_action.cpp:383`). Fields default to `-1.f` = unset; the inline getters fall back to the live globals (`g_ai_aim_min_speed` / `g_ai_aim_min_angle` / `g_ai_aim_max_angle` / `g_aim_predict_time`) whenever a field is negative. An unset NPC follows the `ai_aim_*` console commands as before, and REDONE Combat AI, which drives these 4 cvars, is unaffected. A negative component reverts that component to the global. The sentinel cannot collide with real values since the cvar domains are non-negative.
- `npc:set_vision_speed(factor)` — per-object float on `CVisualMemoryManager`, multiplied into both `get_visible_value` returns alongside `g_ai_vision_speed_boost` (`visual_memory_manager.cpp:368,378`; the cvar stays live: final = base * factor * boost). 1.0 = vanilla, higher spots faster, negative clamps to 0. Works for any `CCustomMonster`; error-log path for anything else. Deliberately not reset in `reinit()`/`reload()` — the manager is reconstructed when the NPC comes back online, so scripts re-apply on `npc_on_net_spawn`.

## Use cases

The aim and vision knobs are globals on per-frame paths, so a script can only set them server-wide (last writer wins). A per-rank or per-situation reaction curve is impossible without per-frame Lua.

- per-rank reaction curves: a veteran swings onto target faster and spots sooner than a rookie
- temporary per-NPC effects: suppression or blinding slows aim, stimulants speed it
- REDONE-style aim tuning without global writes per NPC per tick

## Performance

No Lua on any per-frame path. The setters run from script at spawn or decision points; the per-frame reads are C++ field reads with one negative compare, and the fallback branch reads the same globals the code read before. `predict_object_position` used to read the global through an `extern`; it now reads the owner's field via `m_object`, which is statically `CAI_Stalker*` (`control_action.h:16`), so no cast was added.

## Constraints

- Not serialized: per-NPC values live on the game object and vanish on offline switch; re-set from script on net spawn (noted in the manifest entries).
[...]

PR #606, opened 2026-07-15
Per-NPC best-cover re-pick hysteresis (set_cover_hysteresis)

damiansirbu:
## Summary

NPCs constantly re-pick their best cover point and walk to the new one, which reads in-game as aimless lateral strafing and as flanks that restart before completing. The cause is `CAI_Stalker::update_best_cover_actuality` (`ai_stalker_cover.cpp:246`): the current cover is invalidated whenever a candidate scores `>= current_value + 1.f` — a hardcoded margin. On invalidation `best_cover()` re-picks and `CStalkerCombatPlanner::on_best_cover_changed` resets `InCover`, `LookedOut`, `PositionHolded`, and `EnemyDetoured` to false. As the enemy and teammates move, marginally-better covers keep clearing the `+1.0` bar, so the NPC re-takes-cover every few seconds; and because `detour_enemy` requires `LookedOut` and `PositionHolded` true, every re-pick throws a flanking NPC back through `look_out -> hold_position`.

This adds a per-NPC override of that margin: `npc:set_cover_hysteresis(value)`. Raising it holds a cover through marginal score gains, so an NPC stops strafing between near-equal covers and a flank commits. Sentinel `-1` (default) keeps exact vanilla `1.f`, so unset NPCs are byte-identical.

It follows the existing `set_aim_params` pattern (#594): a sentinel field on `CAI_Stalker` set once from Lua, read in C++. A field rather than a Lua functor is deliberate — `best_cover()` is called from inside every cover action's `execute()` (`stalker_combat_actions.cpp:389,588,1155,...`), i.e. every planner update while the NPC is in a cover action, so a functor at the actuality check would be a per-frame Lua bridge call. A field read is free.

The enemy-proximity invalidator in the same function (`:237`, enemy within `MIN_SUITABLE_ENEMY_DISTANCE` of the cover) is intentionally left alone, so a cover that genuinely becomes compromised still relocates regardless of the hysteresis; the field only damps the marginal-gain oscillation.

## Use cases

- An AI mod that wants "commit to cover once taken" behavior: raise the hysteresis for NPCs it judges to be in a good spot (seeing and firing), reducing the cover-to-cover strafe.
- Stabilizing the engine flank (`detour_enemy`): fewer cover re-picks means fewer `LookedOut` / `PositionHolded` resets, so a flank once started is more likely to finish.

## Files changed

- `src/xrGame/ai/stalker/ai_stalker.h` - field + setter/getter decls
- `src/xrGame/ai/stalker/ai_stalker_inline.h` - setter + getter (sentinel fallback to 1.f)
- `src/xrGame/ai/stalker/ai_stalker.cpp` - field init to -1.f in the ctor
- `src/xrGame/ai/stalker/ai_stalker_cover.cpp` - read `cover_actuality_hysteresis()` in place of `+ 1.f`
- `src/xrGame/script_game_object.h` / `script_game_object_inventory_owner.cpp` - CScriptGameObject wrapper
- `src/xrGame/script_game_object_script3.cpp` - `.def("set_cover_hysteresis", ...)`
- `gamedata/scripts/lua_help_ex.script` - manifest entry

Builds clean on `all-in-one-vs2022-wpo` (DX11, 0 errors).

## Lua usage

```lua
-- pin this NPC's cover: require a candidate 6.0 better before re-picking
npc:set_cover_hysteresis(6.0)
-- revert to vanilla (re-pick on +1.0)
npc:set_cover_hysteresis(-1)
```


damiansirbu:
Closing after deeper testing - this does not achieve its goal. In combat the best_cover() re-picks are driven by on_enemy_change and the on_danger_location_add/remove handlers, which set m_best_cover_actual = false directly. update_best_cover_actuality early-returns at the top when the flag is already false, so the +1.f threshold this PR parameterizes is never reached during a fight, and raising it has no effect on the re-pick rate. The gate only decides a re-pick when the cover is otherwise stable, which combat rarely is. The correct lever is a per-NPC lock at the entry of best_cover(), above all the invalidators - will open that separately.

PR #612, opened 2026-07-18
Fix CTD on ammo sections missing k_ap / cartridge keys

damiansirbu:
## Summary
Makes the 5 cartridge float keys (`k_dist`, `k_disp`, `k_hit`, `k_impulse`, `k_ap`) optional in both ammo loaders, defaulting to the engine's own `SCartridgeParam::Init()` neutral values (1.0 for the multipliers, 0.0 for `k_ap`). Today both `CWeaponAmmo::Load` and `CCartridge::Load` strict-read them via `r_float`, so any loaded ammo section missing one key is an unavoidable crash the moment anything spawns it (loot, trade, `alife_create`) or a weapon chambers it through its `ammo_class`:

    FATAL ERROR
    ...Can't find variable k_ap in [ammo_dyno]

Observed in the wild on `ammo_dyno` (a packed dormant section shipping without `k_ap`). The crash class hits any modpack carrying one malformed or partial ammo section; nothing script-side can guard it because the read happens inside the spawn/chamber path.

Both loaders need the change: fixing only `CWeaponAmmo::Load` leaves the identical strict reads in `CCartridge::Load` reachable through magazine fill.

## Behavior
Identical for every valid section (all shipped sections carry the keys). A section missing a key now loads with the engine's neutral default instead of crashing: missing `k_ap` = no armor penetration, missing multipliers = 1.0, exactly what `SCartridgeParam::Init()` already initializes. The same function already treats `tracer_color_ID`, `k_bullet_speed`, `k_cam_dispersion` and `k_air_resistance` as optional; this extends the existing pattern to the remaining five.

## Tested
Built locally on all-in-one-vs2022-wpo (DX11, 0 errors).

## Files changed
- src/xrGame/WeaponAmmo.cpp: READ_IF_EXISTS with Init() defaults in CCartridge::Load and CWeaponAmmo::Load

### DIFF
--- a/src/xrGame/WeaponAmmo.cpp
+++ b/src/xrGame/WeaponAmmo.cpp
@@ -27,13 +27,16 @@ void CCartridge::Load(LPCSTR section, u8 LocalAmmoType, float ap_mod)
 {
 	m_ammoSect = section;
 	m_LocalAmmoType = LocalAmmoType;
-	param_s.kDist = pSettings->r_float(section, "k_dist");
-	param_s.kDisp = pSettings->r_float(section, "k_disp");
-	param_s.kHit = pSettings->r_float(section, "k_hit");
+	// Optional reads with the SCartridgeParam::Init() neutral defaults, mirroring CWeaponAmmo::Load
+	// below - this site is reached through a weapon's ammo_class (magazine fill / chambering), so
+	// both loaders must tolerate a section missing a key or the CTD stays reachable.
+	param_s.kDist = READ_IF_EXISTS(pSettings, r_float, section, "k_dist", 1.0f);
+	param_s.kDisp = READ_IF_EXISTS(pSettings, r_float, section, "k_disp", 1.0f);
+	param_s.kHit = READ_IF_EXISTS(pSettings, r_float, section, "k_hit", 1.0f);
 	//.	param_s.kCritical			= pSettings->r_float(section, "k_hit_critical");
-	param_s.kImpulse = pSettings->r_float(section, "k_impulse");
+	param_s.kImpulse = READ_IF_EXISTS(pSettings, r_float, section, "k_impulse", 1.0f);
 	//m_kPierce				= pSettings->r_float(section, "k_pierce");
-	param_s.kAP = pSettings->r_float(section, "k_ap");
+	param_s.kAP = READ_IF_EXISTS(pSettings, r_float, section, "k_ap", 0.0f);
 	param_s.kAP *= ap_mod;
 	param_s.k_cam_dispersion = READ_IF_EXISTS(pSettings, r_float, section, "k_cam_dispersion", 1.0f);
 	param_s.u8ColorID = READ_IF_EXISTS(pSettings, r_u8, section, "tracer_color_ID", 0);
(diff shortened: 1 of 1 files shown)


damiansirbu:
Ran in extended play as part of a combined local build (2 installs, 30+ save loads, heavy spawning/loot churn including ammo spawns into inventories): zero cartridge-key CTDs and no behavior change on well-formed sections - the READ_IF_EXISTS defaults only engage on sections missing keys, confirmed by code path on both loaders.

ravenascendant:
the issue identified in ammo_dyno is actually a bug in ammo_base and can be resolved with a simple DLTX file, contrary to the assertion that nothing mod side can be done,  or an update to the ammo ltx file to add the missing key(s) to the ammo_base section.   both are within the scope of an ME bug fix.  

x-ray has a section inheritance system in its ltx system to handle default values.  by having default values defined in a base item section two things are accomplished.

1. exposing the standard shape of an item section of that type to anyone reading the ltx files
2. allows for defaults to be changed in the ltx

having default values in the ltx read in the engine hides the defaults and possibly the existence of the field from modders. It also hides the impact of a malformed item section.  

In general having malformed item sections crash on item spawn is acceptable because anyone adding a section should test it.  

the difference in this case is that there is a bug in an anomaly section and a modder could be forgiven for assuming anomaly sections are correct.  However that doesn't change the fact that correct fix is to make sure anomaly's sections are correct.

themrdemonized:
Removing checks will encourage bad practices and subsequently other PRs to make other fields "optional". Better to actually fix the issue

damiansirbu:
Agree that a DLTX patch adding k_ap to ammo_base prevents this crash through inheritance.

However
- A workaround doesn't make the crash correct. Data correctness and loader robustness are separate concerns.
- Strict reads aren't this loader's convention. It already treats tracer_color_ID, k_bullet_speed, k_cam_dispersion, and k_air_resistance as optional. This simply extends that behavior to five related fields. 
- Failsafe is good but does not help in this case it won't 100% happen on the modder's PC but in a player's save

Tbh I ended up looking at this because I have this destiny of getting reports about any crash, and it's not a rare crash at all nowadays. Whether people use janky mods is a separate issue but still.

It's np, not a big deal anyway.

ravenascendant:
> * A workaround doesn't make the crash correct. Data correctness and loader robustness are separate concerns.
> * Strict reads aren't this loader's convention. It already treats tracer_color_ID, k_bullet_speed, k_cam_dispersion, and k_air_resistance as optional. This simply extends that behavior to five related fields.

ammo_base is not the work around. it is how defaults are intended to be handled. I am 99% sure that the fields you mention are ones that were added post CoP. Before DLTX any mod that messed with ammo would replace the ammo ltx file and would undo any changes made to ammo_base to add new defaults.  so the non strict read was a way to add the fields to the engine without breaking existing mods. 

DLTX changes that in two ways: 
first it reduces the number of mods that directly change the ammo ltx so that is less of a concern.

and second can use DLTX to correct ammo_base in a way mods will not undo by acident until such a time as a major rebase of the game files make sense. IE a new anomaly version or successor project.  

This keeps the original CoP behavior of  no (or few) defaults in the engine, which is still the majority behavior for reading from ltx as I recall. 

There is an argument to be made that x-ray putting all defaults into template sections like ammo_base and crashing on a missing filed isn't the best way to do things. If we want to move away from that it should be singular, or at least deliberatly staged, effort, not a field here and a field there.

PR #613, opened 2026-07-18
Bind make_enemy_visible (force seen-class enemy memory for stalkers)

damiansirbu:
## Summary
Adds `npc:make_enemy_visible(enemy)`, binding `CMemoryManager::make_object_visible_somewhen` - the call the engine itself uses when the agent manager distributes wounded targets across a squad (agent_enemy_manager.cpp). It force-plants `enemy` as a VISIBLE-memory object on the stalker (`visual().add_visible_object(enemy, .001f, true)`), preserving the squad-mask visibility bookkeeping exactly as the engine-internal caller does. Visibility then decays through the normal memory rules.

## Use cases
Enemy selection (`CEnemyManager::evaluate`) scores a currently-seen enemy far above hit or sound memory, so an NPC being shot point-blank by an unseen attacker keeps engaging a distant seen enemy - scripts can already inject the shooter into enemy memory (`enable_memory_object` + `register_in_combat`) but cannot cross that ranking gap. With this bind, a disclosure/retaliation mod puts the shooter into the seen class, where the engine's own nearest-seen logic selects him: the victim and squad actually turn on the shooter. No engine behavior changes without a script calling it.

## Safety
Same guard pattern as the neighboring `register_in_combat` bind: stalker receiver checked by smart_cast, nil and non-CEntityAlive arguments rejected with a script_log error, no-op otherwise. The wrapped function is engine-proven (called internally by the agent manager on live squad members).

## Tested
Built locally on all-in-one-vs2022-wpo (DX11, 0 errors).

## Files changed
- src/xrGame/script_game_object.h: declaration next to register_in_combat
- src/xrGame/script_game_object_inventory_owner.cpp: guarded wrapper
- src/xrGame/script_game_object_script2.cpp: .def next to register_in_combat
- gamedata/scripts/lua_help_ex.script: manifest entry in the game_object Stalker block

## Lua usage
    victim:make_enemy_visible(shooter)  -- shooter enters the seen class; selection follows


damiansirbu:

damiansirbu:
Consumer-tested in extended play (combined local build with my other open PRs, 2 installs, 30+ save loads): a hit-disclosure consumer called the bind live in NPC-vs-NPC fights (squad members receiving a shooter into seen memory, promotions logged per event), enemy selection picked up the promoted target as intended, and dedicated probing of NPC memory afterward showed exact position tracking (memory error 0.0m) with no stale-ghost artifacts attributable to the injection. The nil/dead-entity guards were exercised (mixed squads with casualties) with no errors.

PR #617, opened 2026-07-20
Script Driven Freecam [MT]

TheLostInPlace:
## Free camera lua exports

Adds a script-driven free camera as an `AbsolutePositioning` cam effector
(`SetHudAffect(false)`, so the viewmodel and hud camera stay untouched), plus
five global lua exports:

- `freecam_set(px, py, pz, dx, dy, dz)` -take over the camera; absolute position + direction, zero roll
- `freecam_release()` -hand the camera back to the game
- `freecam_active()` -true while the free camera is driving the view
- `freecam_set_fov(deg)` -FOV override while active; `<= 0` clears, clamped 5–180 (same bounds as the `fov` cvar)
- `freecam_set_roll(deg)` -roll applied to the camera basis, reset on release

The effector is owned by the actor's camera manager (no cached pointers, safe
across level changes), logs `[freecam] activated/released`, and touches nothing
until a script calls it. Exports are documented in `lua_help_ex.script`.

New files: `xrGame/freecam.h`, `xrGame/freecam.cpp`. Registered in
`console_registrator_script.cpp`; effector id `eCEFreeCam` in `CameraEffector.h`.
Built green (DX11-AVX x64/DX-11 x64) on this base.

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -374,6 +374,12 @@
         void set_cam_custom_position_direction(Fvector position, Fvector direction)
         void remove_cam_custom_position_direction()
 
+        // Fov override in degrees, clamped to the fov cvar range, applies to the active FPCam
+        // remove_cam_custom_position_direction also resets this since it destroys the effector
+        void set_cam_custom_fov(float fovDeg)
+        void remove_cam_custom_fov()
+        bool is_cam_custom_active()
+
         // Removes all engine-spawned per-animation HUD camera effectors
         // Scripted effectors added via add_cam_effector are not touched.
         void remove_hud_motion_cam_effectors()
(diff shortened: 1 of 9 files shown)


themrdemonized:
How its different from existing `level.set_cam_custom_position_direction`?

TheLostInPlace:
Fair question, there's more overlap than the PR framing implied. 

`set_cam_custom_position_direction` already covers absolute pos+dir, roll via the bank component, a paired remove, and frame-independent smoothing, which mine doesn't have. So `freecam_set`/`release`/`set_roll` largely re-tread it, the differences are ergonomics (look vector + internal zero-roll basis instead of caller-supplied H/P/B) and effector ordering (absolute-positioning, processed last, so bob/sway can't perturb it; and it doesn't hide the viewmodel/legs like `m_FPCam` does, for a gunsmith-style camera I want the gun visible). 

The genuinely new capability is just the FOV override (`FPCam` never touches `info.fFov`) and an active-state query. Happy to narrow the PR to those two on top of your existing cam instead of landing a parallel system if you'd prefer that.

themrdemonized:
yea, please extend `level.set_cam...` functionality

ljorge705 (approved):

PR #620, opened 2026-07-20
Add an option "bullet_check_visual" in object section to validate bullet hit.

GhenTuong:
Add an option "bullet_check_visual" in object section to validate bullet hit.
Bullets only hit a bone when hitting both bone hitboxes and passing a triangle on the visual model.
Allow a more accurate bullet hit validation. Used for complex objects like physic objects, cars, stationary machine guns.
The bone hitbox must wrap all vertexes of the bone inside.
Not recommended to enable for stalkers, mutants.


themrdemonized:
add new ltx fields and lua exports into ltx_help and lua_help files respectively

PR #631, opened 2026-08-06
Expose script control on whether to activate item on move_to_slot or not

lulnope:
To retain compatibility with existing scripts the boolean flag is flipped - `false` means to preserve old behavior meaning DO activate, `true` means to DO NOT activate the weapon

proof that the changes work correctly is in the attached video:
- equip pistol through equipment UI and it gets activated
- equip pistol through script `move_to_slot` with `true` as 3rd param and pistol is in slot but it does not activate on it's own
- equip pistol through script `move_to_slot` without 3rd param: pistol is in slot and it gets activated so old behavior is preserved and changes are compatible with existing scripts

https://github.com/user-attachments/assets/5f87d4ae-b590-4497-b9e0-5d54dd46b425


themrdemonized:
Add new override into lua_help_ex.script

lulnope:
done, override added to lua_help_ex.script

PR #632, opened 2026-08-10
New Cvars scale_hud_motion_marks_by_speed and allow_accel_during_lookout

Verdatim25:
Adds two new console variables, scale_hud_motion_marks_by_speed and allow_accel_during_lookout.
scale_hud_motion_marks_by_speed is a boolean (default on) that scales motion marks based on the anim speed in the omf, "m_current_motion_def->speed", and speed by either ltx or script changes. I created a new "final_anim_speed" variable for the struct "attachable_hud_item" to store the final_anim_speed so we can use it later for motion mark scaling
allow_accel_during_lookout is a boolean (default off) that allows the actor to accelerate (run) while in the lookout (lean) states. It does not allow sprinting, but allows both walking and running during this state.
Both Cvars dont have russian translations yet, because i dont speak russian and the results i got from google translate were unsatisfactory. so sorry about that

### DIFF
--- a/gamedata/configs/text/eng/ui_mm_modded_exes.xml
+++ b/gamedata/configs/text/eng/ui_mm_modded_exes.xml
@@ -837,8 +837,8 @@
 	<string id="ui_mm_modded_exes_gameplay_weapons_g_auto_reload">
 		<text>Automatically reload weapon when ammo is depleted (g_auto_reload)</text>
 	</string>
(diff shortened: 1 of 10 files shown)


Verdatim25:
just want to put this here, but a fair warning should be issued to modders that have motion marks for animations that have non-1 anim speeds since this will change the timing of them. Although it will cause inconvenience to existing modders, i do believe it is the best course of action to make motion marks more reliable

Issue #445, opened 2026-02-26
LuaJIT 2.1 64 bit version savefile incompatibility workaround


jrowa2:
This issue is going off of the note that is currently in README.md: "**_Future MT versions will include LuaJIT 2.1 64 bit version, it will be incompatible with existing savefiles so for now its on a hold._**"

I am not familiar with the xray-monolith source code, so forgive me if I am suggesting anything that is not possible - this is just question/discussion - I am not sure on how or if this can be implemented. Or if this upgrade to LuaJIT 2.1 64 bit version will be exclusive to xray-monolith.

I suspect this has been considered by people already... I could not see a duplicate issue for this.
Either way, it would be good to make the transition for users as smooth as possible, if playing they are playing on saves prior to this.

What exactly is the issue that would cause the incompatibility with saves prior to LuaJIT 2.1 64 bit? I would assume that something changes with the savefile format.

If the savefile format is changing like I naively suggest... The new savefile contents could contain an extra special flag, or maybe begin with magic header bytes representing if it is post-JIT-update and if not then it requires some parsing and/or re-arranging of contents to then create a savefile that is compatible. That is, assuming that we can do so and "translate" the savefile contents to whatever it is that LuaJIT 2.1 64 bit requires. I would think that the length in bytes and/or endianness of values inside the savefile contents changes.

I would be willing to write code/proof-of-concept for this.

themrdemonized:
The primary issue is with marshal library that performs (de)serialization from Lua into savefiles, the data is incompatible. I personally don't want to waste time to write a code that performs one time only "migration" operation, and then it will be effectively dead for the rest of time. People can start a brand new save when update arrives

PR #641, opened 2026-08-20
feat(luajit): low-overhead CPU and allocation profilers (jit.profile, jit.allocprof)

damiansirbu:
## Summary

Backports the LuaJIT 2.1 jit.profile timer sampler into this 2.0.4 build and adds jit.allocprof, an allocation profiler built on the same stack walk. Both sample the running Lua stack, on a timer for CPU and on a byte interval for allocation, instead of instrumenting calls. One capture attributes cost across the entire script layer with the JIT on, at the sample rate, with no per-function wrapper and no module selection.

jit.profile (CPU): start(mode, cb) arms a timer on a dedicated thread. Each tick latches the VM state and sets the profile hook together with LUA_MASKLINE, so the sample resolves at the next interpreted bytecode, or at the next trace exit if a JIT trace is running (lj_profile.c). dumpstack returns "name@short_src:linedefined" per frame, starting at the executing frame. The module is the upstream 2.1 jit.profile with one structural adaptation for 2.0.4: the stack dump writes into a fixed buffer instead of lj_buf, which 2.0.4 lacks, and there is no prof_mode.

jit.allocprof (allocation): the allocator funnels lj_mem_realloc and lj_mem_newgco add allocated bytes to a counter (lj_gc.c, two integer adds behind an active flag). At the instruction dispatch boundary the drain walks the same stack jit.profile uses and credits the accumulated bytes to the leaf function and the full stack (lj_allocprof.c). Each sampling interval is drawn from an exponential distribution with an 8 KB mean, the method Go's and jemalloc's heap profilers use for unbiased allocation sampling, so a periodic allocation pattern cannot align with the interval.

## Use cases

Anomaly runs hundreds of mods in one shared Lua state. It had no engine-level profiler: CPU cost and allocation were not attributable to a function, script, or callback without instrumenting every call, which taxes the paths it measures and cannot measure allocation at all.

These two modules make script cost measurable. One capture attributes CPU time and allocated bytes to the function, script, and callback that produced them, across every loaded mod, with the JIT on. A modpack maintainer can locate the real cost drivers and optimize by measurement. The allocation axis is new to the platform: GC churn is a primary source of frame-time stutter, and no existing script-side tool could measure it.

## Save compatibility

The full LuaJIT 2.1 upgrade is on hold in this fork because GC64 changes the save and bytecode byte format (#445). jit.profile does not depend on GC64. This backport takes that module alone: no 64-bit GC pointers, no change to the save or bytecode layout. A save written before the change loads after it, and the reverse. The profiler the 2.1 upgrade would bring is available here without the save-format change that put the upgrade on hold.

## Design

Both profilers are inert until a script calls start. When off, the allocator path reads one flag under LJ_HASPROFILE and continues. The engine holds no state and takes no gameplay action. Aggregation, policy, and reporting live in Lua.

Trace recording never starts while a profiler hook is armed (the profiler dispatch branch bypasses the recorder feed), and jit.allocprof aborts any in-progress recording when it starts or resumes. This makes both profilers safe to run with the JIT on: existing compiled traces keep running, and the recorder can never ingest an instruction stream with profiler-induced gaps.

One profiler runs at a time. Both drive the HOOK_PROFILE dispatch bit, so starting one while the other runs raises a Lua error.

## Performance

jit.profile samples on a timer thread at the mode interval, default 10 ms, with the JIT on. The cost is set by the sample rate. There is no per-call instrumentation tax. It is suitable for a long capture during normal play.

jit.allocprof is a session mode: it traps every bytecode instruction while active. Per instruction the drain reads a counter and compares it against the next sample distance. The stack walk runs once per 8 KB allocated on average. It is heavier than jit.profile and intended for a focused capture.

The aggregation tables are static engine memory: 10.65 MB, resident after the first capture of a session.

## Limitations

- Interpreter-anchored, no prof_mode. A CPU sample taken while a JIT trace runs resolves at the next trace exit, so the JIT-compiled share of the VM-state split is a floor. prof_mode requires JIT-backend codegen and is out of scope here.
- No new traces compile while a capture runs (recording is blocked while a profiler hook is armed). Traces compiled before the capture keep running.
- jit.allocprof with the JIT on is safe but coarser: allocations inside compiled traces are credited to the next interpreted frame. Run jit.off() for exact attribution. Total bytes are exact either way.
- jit.allocprof attributes at the instruction dispatch boundary, not the allocation call site. Bytes allocated inside a C call are credited to the next Lua frame, so a high-frequency function absorbs some ambient allocation. The by-root and by-script aggregations carry the reliable magnitude.
- A Lua error inside the profile callback is fatal, matching upstream: the panic handler runs and the process exits.
- debug.sethook line and count hooks are starved while a capture runs. Their hook state is saved on start and restored on stop.

## Files changed

New:

| File | Lines | Purpose |
|---|---|---|
| lj_profile.c | 387 | the 2.1 timer sampler, native dumpstack |
| lj_profile.h | 24 | declarations |
| lj_allocprof.c | 214 | allocation counter, drain, aggregation |
| lj_allocprof.h | 36 | declarations |

Edited:

| File | Change |
|---|---|
| lib_jit.c | jit.profile and jit.allocprof module bindings (+174/-1) |
| lj_arch.h | LJ_HASPROFILE, per-target profile timer selection (+16) |
| lj_dispatch.c | 2-arg lj_dispatch_update, HOOK_PROFILE dispatch branch, profiler-locked hook transitions in callhook (+33/-3) |
| lj_dispatch.h | lj_dispatch_update signature (+1/-1) |
| lj_gc.c | two allocator-funnel byte adds (+11) |
| lj_obj.h | HOOK_PROFILE = 0x80 (+1) |
| lj_state.c | profiler stop on lua_close, call-site ripple (+9/-1) |
| lj_trace.c | no recording start while a profiler hook is armed, call-site ripple (+6/-6) |
| luajit.h | profiling API declarations (+9) |
| ljamalg.c, luajit.vcxproj | build wiring (+4) |
| gamedata/scripts/lua_help_ex.script | manifest entries (+15) |

## Lua usage

```lua
-- CPU: sample every 5 ms
local profile = require("jit.profile")
profile.start("i5", function(thread, samples, vmstate)
    local stack = profile.dumpstack(thread, "", 12)
    -- aggregate stack by count
end)
-- run the scene
profile.stop()

-- Allocation: jit.off() recommended for exact attribution
local allocprof = require("jit.allocprof")
jit.off()
allocprof.start(12)
-- run the scene
allocprof.stop()
jit.on()
local rows = allocprof.dump(true)   -- { {key="leaf;..;root", bytes=N}, ... }
```

## Testing

Built locally on DX11, 0 errors. Deployed to a GAMMA install, demonized MT-based local build, and exercised in play. Also exercised with captures on clean Anomaly 1.5.3 and an EFP install.

jit.profile: a 132 s capture on a live GAMMA session produced 22062 samples across 1138 distinct stacks. VM-state split: JIT-compiled 0.6%, interpreter 6.2%, C and engine calls 63.9%, GC 28.6%, JIT compiler 0.6%. Frame rate during the capture matched play without it, since the JIT stays on. The top cost driver by root frame was the per-frame actor update dispatch at bind_stalker.script:207 at 40.5% of samples, which is also the top leaf at 35.1%.

jit.allocprof: a 139 s capture attributed 1597.5 MB across 1851 stacks with no overflow (the dropped-bytes counter stayed at 0). Top allocators by leaf: the time_global wrapper at _g.script:760 (295 MB, a high-frequency function absorbing ambient allocation per the dispatch-boundary limitation), an item-attachment script (229 MB), and the sorted-iteration helper spairs at _g_patches.script:802 (136 MB). No crash across multi-minute captures.

Extended play: the profiler-enabled exe has run as the normal GAMMA build for two weeks, across many sessions, level transitions, and save-load cycles. With both profilers off it matches the stock exe in play. No crash and no save-load failure over the period.

Save compatibility, both directions: existing stock-exe saves load and continue on the profiler build (two weeks of play). A save written on the profiler build loads and plays on the stock exe (demonized single-thread, profiler removed, same base). The save and bytecode format is unchanged.

Attached: SpeedScope Left Heavy and Sandwich views of the CPU and allocation captures above, and a console shot of a capture session showing the second profiler refused while the first runs.

02-cpu-sandwich
03-alloc-left-heavy
04-alloc-sandwich
05-console-session
01-cpu-left-heavy

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -133,6 +133,21 @@
         table.random(t)
     }
 
+    jit.profile { // profiler-enabled build only; backported LuaJIT 2.1 timer sampler, GC64-free so saves stay compatible
+        jit.profile.start(string mode, function(thread, int samples, string vmstate)) // mode "iN" = sample every N ms; vmstate one of N/I/C/G/J; errors if jit.allocprof is running
+        jit.profile.stop()
+        string jit.profile.dumpstack(thread, string fmt, int depth) // "name@src:line;..." leaf-first; fmt accepted but ignored (fixed format)
+    }
+
+    jit.allocprof { // profiler-enabled build only; per-function allocation attribution, jit.off() recommended for exact attribution
+        jit.allocprof.start(int depth = 1) // byte-sampled (Poisson interval); depth>1 also captures full stacks for a flamegraph; errors if jit.profile is running
+        jit.allocprof.stop()
+        table jit.allocprof.dump(bool stacks, bool reset) // {{key=,bytes=},...} + dropped/pending/drains/seen; reset clears tables but keeps capturing
+        int jit.allocprof.usage() // 0..100 fill of the fullest aggregation table, snapshot before it reaches 100
+        jit.allocprof.pause() // disarm the hook to build a snapshot report at full speed
+        jit.allocprof.resume() // re-arm and continue capturing
+    }
+
     globals {
         string dik_to_keyname(number dik, boolean localize)
         int get_modded_exes_version() // Returns modded exes version in integer format
--- a/src/3rd party/luajit-2/src/lib_jit.c
+++ b/src/3rd party/luajit-2/src/lib_jit.c
@@ -28,6 +28,12 @@
 #include "lj_vm.h"
 #include "lj_vmevent.h"
 #include "lj_lib.h"
+#if LJ_HASPROFILE
+#include "lj_gc.h"
+#include "lj_trace.h"
+#include "lj_profile.h"
+#include "lj_allocprof.h"
+#endif
 
 #include "luajit.h"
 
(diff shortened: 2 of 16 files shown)


Skyki15:
This is a must have for Modders & Modpack owners.
I hope this gets merged 😁

PR #651, opened 2026-08-29
Vendor oneTBB 2023.0.0 (headers, import libs, DLLs)

noisethanks:
Vendor oneTBB 2023.0.0

Replaces the currently-vendored TBB 2020.0 (classic, pre-oneAPI-rebrand)
with oneTBB 2023.0.0, the current stable release. Headers move from
src/3rd party/tbb/ to a combination of a thin forwarding layer there plus
the real implementation under the new src/3rd party/oneapi/tbb/ - this is
how oneTBB itself organizes its public headers, not a restructuring choice
made here.

Also included, since they were dependent on/adjacent to this swap:
- tbb.dll/tbb.lib replaced with tbb12.dll/tbb12.lib + a compatibility
  tbb.lib alias (oneTBB renames the core binary; "12" is a binary-interface
  version, unrelated to the TBB release version). tbbmalloc.dll/.lib and
  tbbmalloc_proxy.dll/.lib added alongside - not currently used by any code
  (nothing in this PR wires them in), but this is the natural place to
  vendor them since they ship as a matched set with the core library.
- src/3rd party/serial/ removed - a 2-file directory that existed only
  because TBB 2020.0's own parallel_for.h referenced it as a fallback;
  oneTBB's version doesn't, leaving it orphaned.
- .gitignore fixed so the new import libraries under sdk/libraries/x64/
  actually commit (the existing /**/x64/ rule was silently excluding them,
  which would have broken linking for anyone building from a fresh clone).

Nothing in the existing engine currently references any TBB symbol, so this
verified as a behavior-preserving change: 0 build errors, and the resulting
exe's import table is byte-for-byte identical to current master (36 DLLs,
no tbb* entries either before or after).

The one exception: src/Layers/xrRender/ParticleEffect.cpp has a
pre-existing, currently-dead #include <tbb/...> that now resolves against
the 2023 headers instead of 2020's. This produces a few dozen bytes of
.rdata/.data difference in that one object file - no executable code
changes, confirmed by diffing the compiled output.

This is the first of three PRs (the other two - TBB-backed scalable
allocator, and a PPL-to-TBB scheduler swap in _thread_types.h - are
independent siblings, both based on this branch, not on each other).
They're not showing up as PRs against this repo yet since they depend on
this branch existing here first; linking them for visibility:
https://github.com/noisethanks/xray-monolith/pull/1
https://github.com/noisethanks/xray-monolith/pull/2


noisethanks:
I have a test build working with this PR as well as allocator and scheduler. Here's the branch. [tbb_build](https://github.com/noisethanks/xray-monolith/tree/tbb_build). Build can be found in Releases.

PR #652, opened 2026-08-29
Fix/thread safety bullet particle

noisethanks:
Fix unsynchronized concurrent access: bullet manager and particle system

Found while investigating thread safety in the engine's multithreaded
render/game-logic split. Both fixes are self-contained. Neither depends
on or relates to any TBB work.

1. CBulletManager::Clear() took no lock. UpdateWorkload, Render, and
AddBullet in the same class all correctly guard m_Bullets with m_Lock.
UpdateWorkload runs on GameThread. Render runs on the main thread. This
is genuine concurrent access, on by default via mtBullets. Clear() only
runs on level load or disconnect, so it's not on the hot path, but it
was the one unguarded member of an otherwise consistent trio. Added the
same lock.

Also removed a stale DEBUG-only thread assertion in AddBullet. It
claimed AddBullet is main-thread-only. It isn't: several weapon/explosive
paths and a Lua binding call it off-main today. The matching assertion
in UpdateWorkload was already removed. This restores consistency.

2. Three sites touch Device.seqParallelBeforRender with no lock:
SItem::OnFrame and SItem::~SItem write to it from a worker thread, and
CRenderDevice::on_idle drains it on the main thread every frame. The
existing onframe_lock doesn't cover this, since it's a per-particle-group
member and can't protect a shared, device-level vector. Added a new
lock on CRenderDevice, following the same pattern already used elsewhere
in the codebase for this exact producer/drainer shape.

Found but not fixed: SItem's destructor has a separate, pre-existing,
same-thread hazard if a deferred callback ever destroys an SItem
mid-drain. Not currently reachable and unrelated to the race above.
Flagging it rather than expanding scope here.

Tested across multiple sessions, including firefights with active
particle effects. No crashes, hangs, or behavior changes observed. This
confirms the new locking doesn't break anything. It doesn't prove the
original race ever caused a specific crash. Both were unsynchronized
shared-state access regardless of whether either was ever observed to
fail.

Edit: Upon closer inspection, it seems that the SItem iterator hazard is reachable. Second commit addresses it.


themrdemonized:
Redo this PR to target MT branch, currently this will merge all MT work (2k commits) into main which is not what i want for now

PR #658, opened 2026-09-01
Expose frame count to the shaders.

LVutner:
Add a framecounter. 

Thank you.


LVutner:
This is not for an Average Joe who's unable to copy&paste files.
This is for modders like Ascii, Ken, Doenitz, or I. 

Animating noise with "timers.x*1000" is more than cringe-worthy. 
Let's be better than majority of those kekray forks. 


This commit does not break SSS23.5, 24, nor custom renderers based of Anomaly.

PR #665, opened 2026-09-08
Shader Bus | Lanes between Scripts and Shaders

TheLostInPlace:
## Quick Disclaimer
This is **NOT** a visual mod. It does nothing to the game visuals, and merely exists to augment/"replace" an archaic system that is being used and abused way beyond what the intended scope for it initially was (quick debugging for developers). There has been a lot of contention around which mods "own" the shader_param(s) and which should not. This PR aims to fix this once and for all.

To re-iterate again, this existing system **STILL WORKS**. The hope is to slowly move away from using the `shader_param_N` and `s3ds_param_N` so that mod incompatibility with clashing slots will no longer be an issue.

## Why

Right now, mods pass values to shaders through eight shared console variables (`shader_param_1` to `shader_param_8`, plus the four `s3ds_param` ones). These slots have no names and no owners. Any script can overwrite any slot at any time, and when two mods pick the same slot nothing warns anyone. On GAMMA all eight slots are already in use, and two enabled mods currently write to `shader_param_5` at the same time. A new shader mod has to either take a slot that may already be in use, or modify the engine.

## How it works
- The engine keeps a list of named "lanes". Each lane is one float4 value with an owner.
- A script calls `shader_bus.register(id, owner, description)` and gets back a token. Only the script holding that token can write the lane. Any script can read any lane.
- A shader uses a lane by declaring uniform float4 bus_<id>. When the shader is compiled, the engine finds the lane by name and binds it automatically. Adding a new lane never requires an engine change.
- Writes go into a pending value first. Once per frame the engine copies every pending value into the value the shaders read. A shader therefore always sees all four components from the same write, never a mix of an old write and a new one.
- If a script tries to register an id that a different owner already holds, the game crashes with an error that names both owners and both script files. `try_register` does the same check but returns nil instead of crashing the game.

For debugging there are `shader_bus.list` (which can also show lanes that a shader declares but no script has registered), `shader_bus.stats` (how many times a lane was written, how many times its value actually changed, the frame of the last change, and the last frame a shader read it), `shader_bus.get_pending`, and four console commands: `bus_list`, `bus_get`, `bus_force` and `bus_release`.

The twelve old console commands keep working **exactly as before**. The only addition is a log line, printed once per console execution and per writing script, that attempts to name the script file and line number that still writes to that command. This makes it easy to find what is left to migrate on a big load order (i.e. GAMMA).

Documentation: `SHADER_BUS.md` in the repository root, the module and the console commands in `lua_help_ex.script`, and an entry in the README.

## Performance

Once per frame the engine loops over the lane list, which holds a few dozen lanes on a typical load order, and copies one float4 per lane. Binding a lane to a shader constant is one integer store per bound constant. Shaders pay nothing extra, because the value arrives as an ordinary constant. Code that does not use the bus is not affected, and the old commands do exactly what they always did. The full list of lanes is printed at level load only when the game runs with `-dbg`.

## Testing

I wrote a stress test mod with 53 checks and ran it on a full GAMMA profile. It covers the id, owner and description length limits, every which way a registration can be refused, the one frame delay between a write and the shader seeing it, the change and write counters, the "last bound frame" value, the console hold and release commands, NaN and infinity values, the writer attribution for the old commands (including writes that go through the `exec_console_cmd` wrapper in _g.script), and 64 lanes written every frame for 300 frames with every read checked. All 53 checks pass. The duplicate owner error was verified with a second mod that deliberately registers an id the first mod already owns.

I made a patch for 3DSS 5.01.2 that moves its scripts and shaders onto 13 lanes for example. Regular play and debugging showed no visible issues. The patch is attached as a proof of concept. [3DSS Shader Bus Patch.zip](https://github.com/user-attachments/files/31938095/3DSS.Shader.Bus.Patch.zip)

An optional Shader Bus Monitor mod is in development. It shows the registered lanes and their values on screen while playing. If you would like it included in this PR for completeness, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,10 @@ The original engine is used in S.T.A.L.K.E.R. Call of Pripyat game released by G
   * Allows to modify contents of loaded xml files before processing by engine by utilizing Lua scripts
   * For more information see DXML.md guide.
 
+* Shader bus
+  * Named, owned float4 shader constants that replace the shared `shader_param_N` slots. A script registers a lane by name and writes it through a token, any shader reads it by declaring `bus_<id>`, and the twelve legacy commands keep working while the log names whoever still writes them.
+  * For more information see SHADER_BUS.md guide, the Lua module and console commands are listed in `lua_help_ex.script`.
+
 * Possibility to unlocalize Lua variables in scripts before loading, making them global to the script namespace
   * For unlocalizing a variable in the script, please refer to documentation in test file in `gamedata/configs/unlocalizers` folder
 
--- a/SHADER_BUS.md
+++ b/SHADER_BUS.md
@@ -1,0 +1,251 @@
+# Shader Bus
+
+---
+
+## Intro
+
+The shader bus is a named, owned replacement for the eight anonymous `shader_param_1..8` and
+`s3ds_param_1..4` float4 constants. Any script can register a lane by name, any script can read
+any lane, and only the script that registered a lane can write it. The engine needs no advance
+list of names, so a mod can claim a lane without anything being patched into the exe for it.
+
+Requires a modded exe carrying the bus. `shader_bus.version()` returns `2`. A script that wants
+to detect the feature should test `shader_bus ~= nil` first, since an older exe has no such
+module at all.
+
+---
+
+## The shader side
+
+Declare a `float4` uniform whose name starts with `bus_`.
+
+```hlsl
+uniform float4 bus_myeffect;
+```
+
+That is all. When the pass compiles, the renderer sees the name in the reflected constant table
+and attaches a binder that writes the lane's current bound value on every constant table switch.
+The part of the name after `bus_` is the lane id, so `bus_myeffect` is the lane `myeffect`.
+
+Declaring costs nothing extra to compile and nothing to run if the constant is never read. A lane
+nobody has registered reads `0, 0, 0, 0`, so an unclaimed lane behaves like an unmodded install as
+long as your shader treats an all-zero value as off.
+
+---
+
+## The script side
+
+```lua
+local token = nil
+
+function on_game_start()
+    if shader_bus then
+        token = shader_bus.register("myeffect", "My Mod Name", "what this lane carries")
+    end
+end
+
+function actor_on_update()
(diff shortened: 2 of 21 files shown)


themrdemonized:
I recommend you to translate Claudish descriptions into human readable English, this is hard to read and understand

TheLostInPlace:
> I recommend you to translate Claudish descriptions into human readable English, this is hard to read and understand

Which parts are you having difficulty understanding with? 

If I used Claude here it likely would've have been easier to actually understand probably lol. 

Also just for the record, I have used Claude in the past and when I do, I don't shy away from noting it as such, the reason I formatted it this way and used this language is because this is a semi-large PR so I wanted it "clear" to follow through.

themrdemonized:
> Which parts are you having difficulty understanding with?

From the How section and so on.
Also i think you should target main branch with this one, otherwise there will be discrepancies in how engine loads shader parameters

TheLostInPlace:
> > Which parts are you having difficulty understanding with?
> 
> From the How section and so on. Also i think you should target main branch with this one, otherwise there will be discrepancies in how engine loads shader parameters

No worries, appreciate the feedback. I'll re-write it in a bit. As for the retargetting, do you want me to make a separate PR to MT branch if there's a conflict? Or I guess this could be the MT PR and the other one I'd open with be base.

Going through the rebasing now, looks like there are minimal changes, if any. So I'll just make this target the main branch and it should merge cleanly into MT. The semantic difference is that `Core.isDebug()` doesn't appear to exist in MT so I'm using `strstr(Core.Params, "-dbg")` instead, which appears to be the norm on the main branch.

TheLostInPlace:
Re-based and force pushed. Just tested and the same stress test mod I wrote works perfectly fine on non-mt exes. Let me know if you have any issues.

Additionally, I've uploaded the stress test mod itself. Just load it last in MO2 and make sure it wins combine_2_naa for the visual confirmation (rainbow effect), the actual test results will be in the log.
[Shader Bus Stress Test.zip](https://github.com/user-attachments/files/31951647/Shader.Bus.Stress.Test.zip)

PR #669, opened 2026-09-12
jit.profile / jit.allocprof: hardenings and jit.util.gcstat

damiansirbu:
Resubmit of #663. Its base branch `pr` was deleted, so a reopen was not possible. This is the same content against the current `all-in-one-vs2022-wpo`.

Hardenings on top of the merged #641 profilers, from running them as the backend of an in-game profiling UI on a full GAMMA install:

- allocprof stack keys cut at frame boundaries instead of truncating mid-frame, so a dropped frame drops whole and attribution stays correct
- buffer and depth sizing for deep stacks (dumpstack 2048 to 8192, alloc key 512 to 4096)
- interval jitter so the sampler cannot phase-lock with per-frame work or the 100/200ms script throttles
- heap-allocated aggregation tables, so a session that never profiles holds none of the static image data
- `jit.util.gcstat`: threshold, estimate, debt, and a completed-cycle count, for reading collector health from Lua

## Testing

Single-thread and multi-thread, 2-minute combat captures on a vanilla-based install and on full GAMMA.
0% of samples at the depth-64 cap, the VM-state split reconciled with the per-stack weights, and gcstat coherent across a fight (threshold at or above total, estimate at or below total, debt bounded).

### DIFF
--- a/gamedata/scripts/lua_help_ex.script
+++ b/gamedata/scripts/lua_help_ex.script
@@ -148,6 +148,10 @@
         jit.allocprof.resume() // re-arm and continue capturing
     }
 
+    jit.util { // additions beyond stock jit.util
+        table jit.util.gcstat() // {total, threshold, estimate, debt} in bytes from the LuaJIT GC state; read-only snapshot, values may be mid-cycle under parallel GC
+    }
+
     globals {
         string dik_to_keyname(number dik, boolean localize)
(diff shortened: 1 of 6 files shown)


damiansirbu:
Superseded: moved the branch to a clean fork. Reopening from there.

PR #675, opened 2026-09-16
scripted attachments: lights and 3d ui on attachments, a shader parameter per mesh child, wider debug lines (linked PR #2)

TheLostInPlace:
## Changes

Additions to the script attachment manager and the renderer behind it.

- a script can hang a light or a 3d ui window on a bone of an attached model, under a key it
  picks, and find it again later
- a float4 shader parameter per mesh child on skinned models, written every draw, so one attachment
  can drive shader effects on each of its parts (we use it for the watch tubes)
- setting the same shader and texture again no longer rebuilds the model
- script debug lines get a width and a depth test flag, wide lines draw as quads in world units
- a hud attachment can ask to be drawn even when the hands hold nothing, its ui too

No change for anything that does not use the new calls. Every new export is listed in
`lua_help_ex.script`. Built DX11-AVX x64, tested in game. Merges into the MT branch without
conflicts, but there `render_hud_attachments_always` has to take the dsgraph manager like
`render_hud` does. The MT PR has that.


damoldavskiy:
I kindly ask not to merge this until we're done with scripted attachments rework with the creator of scripted attachments

PR #677, opened 2026-09-16
player hud: scripts can drive the hand animations, show empty hands and offset each hand (linked PR #4)

TheLostInPlace:
## Changes

Script bindings on the player hud so a mod can move the hands without holding an invisible item.
While initially made with Wearable Devices in mind, nothing in it is tied/specific to
that mod.

- reads of the hud animation state, the freelook state, the hud fov and whether the hud is visible
- a position and rotation offset per hand with a blend time, `game.set_hud_offset` and
  `game.clear_hud_offset`, and script motions that keep playing while freelook is on
- a bare hands mode that draws the hands holding nothing, plus reading and setting the hands
  section
- a way to override the movement blend, and a full animation resync for scripts that swap hand
  models
- the hand motion tables are rebuilt when the hands model reloads, so a swapped model does not keep
  the old motions
- with bare hands up the attachment ui keeps drawing, and the frames skipped without hands are
  counted so a script can check
- two callbacks through `callbacks_gameobject.script`: `actor_hud_on_script_anim_end` (part,
  section, motion, reason) and `actor_hud_on_hands_changed` (new and old hands section)

Nothing changes unless a script calls the new functions, the bare hands mode is off by default.
Every new export is listed in `lua_help_ex.script`. Built DX11-AVX x64, tested in game with the
Wearable Devices mod. Merges into the MT branch without conflicts.


themrdemonized:
Marked all parts as drafts for now. Lmk when they are ready

TheLostInPlace:
switching them back to ready, should be no more surprise last minute fixes. anything that comes up not related to this will just be a separate PR, I didn't want to have you rebase for HUD rendering pathway that I was altering anyways

